PR #112, opened 2016-08-21
Nuget cleanup

MaitreDede:
No need to store Nuget packages to repository, Visual Studio will download them at compilation if needed.


antonpup:
Thanks for the tip! I appreciate your help.

PR #244, opened 2016-12-08
AtmoOrb ambilight device support

RickDB:
Tested successfully on master branch and converted over to dev, there were some already present compile errors on the dev branch related to profilename so couldn't compile it successfully because of that.

However checked error logs and AtmoOrb code looks valid returning no errors, first time working with Aurora code so please let me know if you require any changes :)

Just not sure about the config enabled check, it works but would prefer it to spawn or destroy the device on UI toggle because it will report connected state properly however in configuration code don't see an option for it atm.

The new config tab "Amblight" is made to add Hyperion in the same area, that way we don't waste any space with a per ambilight specific tab.


RickDB:
Other than that the color calculation in AtmoOrb isn't ideal, it takes the average of eache RGB value separately which does give the desired results during testing here but want to improve it later.

Tried looking up some examples and how we do it with AtmoLight but those are all bitmap based, to get an average color of the keyboard as a whole I'm not entirely sure how to do that yet.

antonpup:
Take a look at how ClevoDevice handles lighting. They're region based, so I exposed keyboard bitmap for region color averaging. Should be what you're looking for.

antonpup:
I resolved the compile issue.

RickDB:
Nice :)
Playing around with the way Clevo gets the bitmap but not sure if I'm doing this right:

```
            BitmapRectangle keymap_ctrl = Effects.GetBitmappingFromDeviceKey(DeviceKeys.LEFT_CONTROL);
            BitmapRectangle keymap_numMinus = Effects.GetBitmappingFromDeviceKey(DeviceKeys.NUM_MINUS);
            BitmapRectangle regionAll = new BitmapRectangle(keymap_ctrl.Bottom, keymap_numMinus.Top, 128, 128);
            Color regionAllColor = Utils.BitmapUtils.GetRegionColor(colorComposition.keyBitmap, regionAll);
```

So what this should do is get the entire region starting from left control (bottom) to num_minus (top) with a fixed bitmap of 64x64, am I doing this correctly?

antonpup:
My apologies for a long delay. Here's code that will take the average of an entire bitmap.

``` C#
Color bitmapAverageColor = Utils.BitmapUtils.GetRegionColor(
                        colorComposition.keyBitmap,
                        new BitmapRectangle(0, 0, colorComposition.keyBitmap.Width, colorComposition.keyBitmap.Height)
                        );

Global.logger.LogLine("Average color is " + bitmapAverageColor.ToString());
```

The logger line can be removed, I was just using it to view the color.

RickDB:
No problem :)

Thanks for the example code, will resolve merge conflicts and add that in 👍

RickDB:
Everything merged up to current dev branch, compiled and tested locally here without errors but worth double checking of course :)
Not sure about travis check failing though as it was during initial PR.

antonpup:
``` C#
var multiCastIp = IPAddress.Parse("239.15.18.2");
var port = int.Parse("49692");
```

Do these change from configuration to configuration? If so, then add these as text field options in the Orb settings. Also, can you create an "AtmoOrb" tab and place all settings for it there. It's getting kinda crowded in other tabs in settings.

RickDB:
The multicast address info is static so for the Orbs we can leave non-configurable to avoid any issues with users changing them by accident :)

Thought I already added that under a new Ambilight tab instead of other as to also place Hyperion settings in there (saves some room), here's how it looks here atm with some room left on the right for Hyperion:

http://i.imgur.com/Zuz9Lgb.png

Or do you mean in other area then settings there?

antonpup:
Oh, that is fine. I must have missed it.

Thank you for your contribution and helping this project grow! :D

RickDB:
Glad to help out 👍 

Next up is Hyperion support which will bring a lot of led device support (SPI / USB / Networked etc..) at once without bloating the code with per device libraries and what not :)

antonpup:
Could you look into disabling AtmoOrb device if the Orb doesn't respond (doesn't exist). As a concern was raised by a user #342.

RickDB:
Will take look :)

antonpup:
Is is possible that the constant network usage from AtmoOrb can cause high ping as reported on #359

RickDB:
Very minimal packet rate and never seen that from users so far, basically it will send at the rate we send keyboard leds.
But either way those are very small packets locally which shouldn't disrupt anything, looking thru code now and checking why we have network traffic on just the connect :)

But if AtmoOrb is enabled in config it will not know if Orbs are reachable (UDP doesn't report state) so will keep on sending, Hyperion is better at this as it proxies it to all devices.
Have that working just lack of time lately but will get around implementing support for that soon :)

RickDB:
Added configurable delay and made sure we never start the multicast receiver when user doesn't have setting enable in this PR:

https://github.com/antonpup/Aurora/pull/363

Hopefully this will help reduce the load but just let me know if this needs further improvements :)

PR #251, opened 2016-12-11
Half-Screen Screenshot (Posibility of Custom Screenshot Region) [More CPU Friendly] 

rkkoszewski:
I implemented a small alternative to Pranas.ScreenshotCapture.TakeScreenshot which allows to use custom regions for the screenshots for Ambilight. This reduces CPU usage (Previously 15% and now 2-4% with dual 1080p monitors)

Using only the lower half of the screen not only uses less CPU but simulates a glow effect from the bottom of the screen. (Tested on a Clevo Device). Maybe for some more precise keyboards applying some blur kernel to the canvas could give a more convincing glow effect (I can look into that).

Not sure if anybody is interested in this feature, has some other ideas to share, or just hear if this could be something useful to have. I created this pull request in case somebody is.


manskou:
I was trying to do the same thing, only I used autohotkey and python. However, due to my lack of coding skills (i am a beginner), I couldn't create the final script that aurora will use.

antonpup:
I will look over your PR at a later date. I will most likely rewrite Ambilight to include more options like: "bottom of the screen" (how you implemented it), "entire screen", "monitor selection?", "average color instead of scaled image".

I also would like for Ambilight to work with DirectX, OpenCL, OpenGL, whatever have you 3D engine. So users don't have to run games in borderless windowed to enjoy ambilight effects.

rkkoszewski:
Hi @antonpup ,
Have you looked seen the DFMirage Driver?
From what I could find out it has really low CPU usage, and is able to capture from full screen applications (DirectX and co?). Here is the article: http://www.instructables.com/id/Ambilight-DFMirage-super-Fast-Screen-Capture-and-P/

And here is a sample code how to use it in C#: http://www.demoforge.com/sdk/MirrSharp.zip

I didn't have time to look much into it as I'm now in the middle of exams. But I'll try it out as soon I find a moment.

rkkoszewski:
So I've been tinkering around a bit with the DFMirage Driver and capturing the full desktop at around 60FPS (Not 100% accurate) it was barely taking from 3 to 6% of CPU. That's just amazing! 

Still haven't tried if it captures DirectX too (But from what I've read so far it does). Will do a small test implementation in Aurora when I'll find a moment.

rkkoszewski:
Wow.. Well.. This merge request got a bit messed up. Didn't knew it would gather my last commit. 
I'm still loking into the MirageDF driver thing. It gets full screen capture at +-60FPS for the same CPU usage as my half screen implementation, without having that stutter in the windows.

rkkoszewski:
Good news! I managed to implement the MirageDF Driver, and it works! It is a ver quick and un-optimized implementation, to check if it even works. Right now there is quite some lag happening which I believe I can further minimize.

The good thing is that it fully supports DirectX games, and so on. I tested it with Broforce and Watch Dogs 2. There is some downsides tough. It doesn't seem to capture everything 100%. There are some weird transparent pixels, which must be pixels that are filled in afterwards or something, so some elements of the GUI are not rendered, or in the case of Watch Dogs 2, the main character and UI elements appear as transparent. I've never seen a bitmap with transparent color. By now I'm filling it up with black color and it seems ok, but I'll dig into it when I find some more time.

On the other hand it seems that the driver only works when Aurora is compiled in 32 bits (64 bits crashes the driver), but maybe there is some workaround for this. Altough if it's using a 32 bit DLL, not sure if that's even possible to load from a 64 bit environment.

The CPU usage as of now is roughly the same as the current one in Aurora, but having the capability to record full screen Direct X games. But I read that it should be possible to capture at 90FPS with no more than 15% of single core CPU usage (The CPU wasn't specified tough).

Is there any way to speedup the loop or refresh time from Aurora? Is it possible to have Aurora running at 60FPS?

Input on this topic from @antonpup would be highly appreciated.

Greetings.

antonpup:
Well... that looks like a good solution. It's possible that the driver isn't compatible with Windows 10.

As for fps, you don't want to run Aurora on high fps, otherwise, it causes performance issues and you don't see any "additional frames" on your keyboard lights anyway. Aurora runs at 30fps at the moment, which is good enough for effects. So, just cap it at 30 fps, and that will do the job just fine. Does it also capture multiple monitors? And how about capturing specific application instead of entire screen?

Please try to get it working on x64.

antonpup:
Were you able to get it working on x64? Looking to release a new version sometime soon, and would like to include as much as I can.

rkkoszewski:
Hi @antonpup.

I'm pretty sure it works correctly on Windows 10. The implementation from http://www.instructables.com/id/Ambilight-DFMirage-super-Fast-Screen-Capture-and-P/ is using the same C# class (With probably some modifications) and I have no glitches on my pc. I still have to check what he has done to fix the missing pixels.

As for the x64 implementation, I was not able to achieve to make it work. It crashes while trying to access some buffer. I'm not sure if its possible or at least how to (force) load a 32 bit DLL into a 64bit C# executable. It seems to be importing the system's "user32.dll" which should probably be 32bit even when the 64bit version is launched. I don't know how to force load the 32 bit version.

As for multi-desktop support. From what I've seen so far it only supports full screen capturing of the main screen. I didn't have time to dive much into it.

Finally, there is one small drawback that comes from using the DF Mirage driver. When the driver is first initialised, it resets the mouse pointer to the center of the screen. This makes a weird feeling when you have Aurora set to start automatically on system start. While everything is starting up, after Aurora gets started, the mouse suddenly goes to the center, this may be annoying for some people. (But after all they can choose to use it or not).

The capturing efficiency pretty impressive tough. For 30 FPS it uses around 8% of CPU for a full screen capture that supports DirectX, OpenGL, and causes no lag while dragging windows around on the desktop.

I still believe I can make it work, (At least remove the capture glitches). I will try to force use the 32bit DLLs and see if it works like that on 64bit. The only thing is that I probably won't have time to get to it till next month, as I'm right now in the middle of final projects and exams. 

The Arduino RGB Device is working pretty nicely toguh, and the new settings menu for devices is really handy and easy to use. It's just stupid from me that I didn't branch it and it all got now into this one Merge Request.

RobertLadig:
Hi! 
Did something come out of this? It would be awesome to have an Ambilight that only targets certain areas of the screen. I play a lot with gamepad and especially targeting UI element positions and averaging their color over one or several keys would make for some really great custom profiles in some games.

simon-wh:
@RobertLadig This implementation lacks support for x64 which is what Aurora is usually built against. The implementation needs to be running on x64.

PR #269, opened 2016-12-22
Replace filename ' .cs' to '.cs'

jayvdb:


jayvdb:
Thanks for the fixup at https://github.com/antonpup/Aurora/commit/79b48d2dfff39471 . My apologies for not noticing that.

PR #275, opened 2016-12-26
Dim background option in Volume overlay effect

cvb941:
I added an option to black out keys under the volume overlay area before applying the volume overlay effect itself. This way it's better seen on backgrounds where the effect would otherwise blend in, such as animated rainbow or solid green color.

It's working, but inefficiently using a gradient EffectLayer just for a solid black background, I basically just copied the procedure from the volume overlay itself, because I didn't delve into the code much. So maybe it's more appropriate to consider this as an feature request and change this PR to a proper implementation, with different variable nomenclature according to the codestyle etc.

### DIFF
--- a/Project-Aurora/Profiles/Overlays/VolumeOverlay/Event_VolumeOverlay.cs
+++ b/Project-Aurora/Profiles/Overlays/VolumeOverlay/Event_VolumeOverlay.cs
@@ -1,6 +1,8 @@
 ﻿using System.Collections.Generic;
 using Aurora.EffectsEngine;
 using NAudio.CoreAudioApi;
+using System.Drawing;
+using Aurora.Settings.Layers;
 
 namespace Aurora.Profiles.Overlays
 {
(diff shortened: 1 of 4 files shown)


antonpup:
Take a look at the shortcut assistant for dimming background.

https://github.com/antonpup/Aurora/blob/dev/Project-Aurora/Settings/Layers/ShortcutAssistantLayerHandler.cs#L109

cvb941:
Thanks for the pointer, it should be properly done now.

antonpup:
I'll merge it, but I will change the black color to be a bit more transparent.

Thank you for your contribution!

cvb941:
I personally liked the black to be opaque, I would prefer it that way even by the shortcut assistant.. Maybe add an opacity slider next to both?

Thanks is on my part, this was actually my first ever pull request done on github or just some contribution to an open-source project at all. Also, I've never really coded in C#. I'm feeling quite contented now :)

antonpup:
Might as well let people choose the dimming color (with alpha)?

cvb941:
I thought about it, but thought that it wouldn't really be needed. But now that I think about it more, at least white color would be considerably useful. Of course at this point, let's just let people choose any color they want. The only against is, maybe if someone wanted to replicate the original dimming function it would be more hidden and less accessible this way, but I guess people who use this program are 'able' anyway and will have no problem figuring out that it's just black color with lower "A:" value :)

But after that, it would need to be renamed to background color or something else briefer and clearer, since it's no longer just "dimming" the background.

antonpup:
I just added it as "dimming color". I think it's still a valid name, as it starts off as a semi-transparent black color.

The feature is there, I can always change the visual name later.

cvb941:
Allright, sounds good :) 👍

PR #302, opened 2017-01-01
Small fix EffectLayer.PercentEffect didn't use whole color spectrum.

VoronFX:
For example we have 5 keys for displaying percent effect. 
They have indexes 0-4. 
keys.Count() is 5.
Than for 100% we are taking color `(float)i / (float)(keys.Count()`  = 4/5 = 0.8 in spectrum.
Instead we should take `(float)i / (float)(keys.Count() - 1)` = 4/4 = 1 in spectrum


antonpup:
Good catch. Thanks!

PR #335, opened 2017-01-08
New Color Enhancement Method with HSV

LukasACH:
I have added two additional methods to calculate a brighter color for the Razer Wrapper.

- Simple: A simple multiplication with some overflow detection, and
- Gamma: An implementation of an image gamma correction.

The first one isn't very special, but some people might use it.
The second is pretty good a brighten very dark colors, but makes light colors a bit gray.

In addition, I removed the initial_factor because it wasn't needed.


antonpup:
Well done, it works very well. Thank you for your help! :D

LukasACH:
I am very happy that I clould add my part to your awesome project. Thank you too for your work!

LukasACH:
@antonpup I recently saw that the textbox which should give a hint on wich algorithm to use doesn't have the right text in it. It's from a previous state and I must have overlooked it. Could you change it to the following, I don't want to open a new PR?

> HSV is an algorithm to brighten dark colors, while not affecting bright colors in an elegant way. To do this, it needs more calculations per color, therefore uses more CPU power. If you notice any delays, you should consider not using HSV.

The text is found in all three places where the color enhancement is in place, each on the lines 52, 53 and 57.
Thanks for your work!

antonpup:
Done.

PR #340, opened 2017-01-09
Masterkeys Pro L White support

mm615657:
Since Masterkeys Pro L and Masterkeys Pro L White has same layout so I can add it easily .

Still need works on MasterKeys Pro M White support.


mm615657:
I'm not knowing too much on codes so...guess i need someone else for help....

snichola:
Your code looks good.  What do you need help with?

mm615657:
@snichola i tried to add _L_White and _M_White as separate configuration but new layout can't load :/

mm615657:
ahh i know where is the problem.working on it now

mm615657:
It's loaded now.but we need to delete the space taken by the space between command keys and arrow keys...but i don't know how....

mm615657:
Well it is loaded now .But layout for _M_White is wrong,We need to delete the space taken by the space above the arrow keys....i don't know how :/

antonpup:
Before I approve this PR, I would like @snichola to verify that these changes don't affect his keyboard integration. In other words, it fixes bugs and doesn't break previous support.

Also, for those arrow keys. I will look into making them work. But as of now, Aurora isn't designed to handle keys with multiple functions. So they would either be the Numpad keys (probably this) or the arrow keys.

snichola:
My keyboard still works.  I have the Pro S so I can't verify fixes for the numpad and profile keys but they look correct.

mm615657:
Yes,the problem is Numpad keys the M version keeps Numpad but not arrows and command keys,so I removed them at json file.but that is generated from the default layout so the space on the left side of Numpad still there.maybe we should add additional check that if a key defined twice Aurora should load the feature file.

(Or we just redefine all keys and ignore default layout)

snichola:
Sorry I'm a little confused.  Are you having problems with the virtual
keyboard layout or the actual mapping of the Aurora's DeviceKeys to the
physical keyboard?  I ask because I just looked at the pro m coordinates
and realized they won't align correctly.

mm615657:
It is only virtual keyboard layout have problem ,the actual mapping to physical keyboard is fine

snichola:
If that's the case I think @antonpup would be of more help.  Just to reiterate for Anton my implementation still works and it looks like it fixes some incorrectly mapped keys as well.  Btw which keyboard do you have?  I assume it's the Pro L White.

mm615657:
The one currently in my hand is the Pro L White.

antonpup:
I'll merge this later today and fix up the virtual layouts.

But I do have a question. Does Masterkeys support foreign layouts? And if so, have they been included in the support? I'm talking about the extra keys that aren't found on US keyboards, not the layouts, the layouts are managed by Aurora.

mm615657:
this might be help 
qq 20170112220152

snichola:
Do you have a foreign layout?  I haven't been able to find an image of an "EU" masterkeys keyboard.  In any case we can add support for foreign layouts if they exist but I might have to look into a better way of managing the keys.

antonpup:
It was my fault for not checking, but next time, please do pull requests for the dev branch, not master.

antonpup:
I saw the issue with the layout. I will be reworking the way keyboard layouts are modified. Because as of right now, there is no way to adjust the left margin.

antonpup:
I got the layout working, however, I have come upon:
```
"included_features": [
	"masterkeys_pro_m_white_right_features.json"
  ]
```

Which [Masterkeys Pro M White](http://www.coolermaster.com/peripheral/keyboards/masterkeys-pro-m-white/) does not have any extra features. So... I have excluded it. Any reason why it was added in the first place?

PR #349, opened 2017-01-10
Rainbow Six Siege Profile?

MRiGGz:
I've already seen the amazing work you guys have done on the Overwatch profilem, but could we possibly see a profile for another great game (Rainbow Six Siege). It doesn't seem to be that very popular or I could just be seeing things wrong but I do see the potential of making a profile for this game. It doesn't even have to be that advanced it could be something just as small as creating different lighting for different squads in the game i.e (GSG9 squad yellow, Spetsnaz red, GIGN purple, etc.) I would really appreciate it if we could see a possible update in the future featuring this. If its too much effort, could I get some help to make it myself?


antonpup:
What a peculiar way to contact me. Out of all options: Steam, Issue, email, a pull request. Interesting.

Anyways, I have the game, I'm enjoying it. But I am yet to find a way to get information out of the game. There's no logging file, no lighting integrations, nothing. And I don't want to read the screen for information because it is a slow and resource intensive process. So... I don't know if it is possible to get any dynamic lighting for R6S.

Ge0rges:
@antonpup is that still the case?

Issue #342, opened 2017-01-09
Constant Upload


selmansavas:
Aurora contantly uploading with 3-4kb/s, what is it for?

antonpup:
That appears to be the AtmoOrb device. It's constantly trying to maintain Orb lights. I will see about fixing it.

Here's the definition of the IP address that you're probably seeing. https://github.com/antonpup/Aurora/blob/master/Project-Aurora/Devices/AtmoOrb/AtmoOrbDevice.cs#L110

Also note, Aurora's Wrappers and GSI implementations may also show up in the networking, as it's trying to communicate to other apps via IPC (for wrappers) or http requests (for GSI).

selmansavas:
Thanks for fast reply, Great work with Aurora. Do you have a patreon or another type of tip method?

antonpup:
There are options to donate at the [bottom of github project page](https://github.com/antonpup/Aurora#donations).

PR #363, opened 2017-01-11
AtmoOrb - improvements

RickDB:
- Added AtmoOrb send delay in settings (defaults to 50ms).
- AtmoOrb only connects if user has setting enabled.

Should help reduce the network load mentioned in issues #342


antonpup:
I changed the way devices are handled now. Take a look at the latest dev build. Would a solution like that suffice your needs?

PR #367, opened 2017-01-12
Added more options to Key Fade effect

DeathAngel:
Added the option to assign which keys are gonna fade.


antonpup:
Thank you for your contribution. I think I will expand this to other effects, as well as flip the option. So instead of "These key will be affected" it will be "These keys will not have an effect".

Issue #430, opened 2017-02-02
Game Detection doesn't work properly - Fix included


r3sist3nt:
### Issue:
When using custom Profiles, Aurora won't change the Profile when the Game is started.

### Expected behavior:
Profile will change
### Actual behavior:
Profile isn't changing.

### Fix:
I already tracked down the issue to the way aurora is getting the Process FileName.
When accessing games with a x64 Architecture, an Error might occur due to different architectures, even when Aurora is compiled in x64.

There was a similar problem in #415, but i wasn't sure if its the same Problem so i opened a new one.

I fixed it by changing the functions to only use WINAPI calls, so aurora don't need to directly access the Process.
I will open a pull request to dev with the fix.

Issue #424, opened 2017-01-30
ARK Survival Evolved issues


DMZDesolace:
I'm attempting to create a new profile for ARK, but when the game is loaded Aurora won't switch to the profile. Both Foreground App Scan and Windows event options have the same issue.

Note: The game uses Unreal Engine 4 (shootergame.exe), and uses BattleEye anti-cheat.

Issue #147 - Is opened to add Native ARK support, but i believe that the profile not switching is a different issue. Please correct me if i'm wrong, thanks.

antonpup:
ARK is probably run as admin, so try running Aurora as admin as well.

There is a permissions issue where if a normal process attempts to retrieve information about an admin process, it will not allow it to do so.

DMZDesolace:
Thanks for the info.

Running Aurora as Admin, and the game without BattleEye makes the Profile run successfully.
However, if the game is run with the anti-cheat, the profile still will not load.

(Note: i made a separate profile for the anti-cheat executable [ShooterGame_BE.exe] which loads initially, then switches to the Default profile. I checked Task Manager, and the process is still ShooterGame.exe]

Any ideas why this might be? i would expect the Foreground App Scan to work in this case.

antonpup:
It probably means that exe name is reported as something else or anti-cheat is doing some weird permissions magic. It's also possible that battle eye has an overlay, and when you're running the game, it reports exe as battle eye's.

r3sist3nt:
I found a bug in the Process detection with x64 games, ark might be also effected by this. I opened a PR with the fix.
I will test later if it also fixed the issue for ark.

stbatuhan:
Hi, i didn't test it now but you can add extra option for ark. Maybe if someone has a problem then application can search for the active tasks. And if there is Ark then can start that profile. Or if there is active one is battle-eye then app can search for ark or h1z1 or smt else to activate right profile.

PR #432, opened 2017-02-02
Changed the Process detection to WINAPI only.

r3sist3nt:
### Fixes:
#430 
#424

The old Process Detection was using the ProcessModule to get the FilePath. This could cause problems, because games, compiled in x64 can't be accessed in this way.
I changed the detection to only use WINAPI calls, so Aurora don't need to access the Process directly anymore.
### Changes:
I only needed to change the GetActiveWindowsProcessname() Method in GameEventHandler.cs
I'am using the `GetProcessHandleFromHwnd()` call to get the process handle from the foreground window, then i get the full path with the `GetModuleFileNameEx()` call.

No Exception handling is needed, the returned fileName will be an empty string when no foreground window exists.

### DIFF
--- a/Project-Aurora/GameEventHandler.cs
+++ b/Project-Aurora/GameEventHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text;
 using System.Collections.Generic;
 using System.Diagnostics;
 //using System.Timers;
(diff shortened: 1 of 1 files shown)


antonpup:
Thanks for the fixup. I am confused as to why the old method doesn't work properly? Do you have a more detailed explanation for that? Also, Does this allow Aurora (if not run as admin) to retrieve the name of a process running as admin?

r3sist3nt:
I was running into the problem when i tried to creating a custom profile for Rust. Ich tried everything, but ist just doesn't want to work.
The game itself was running with user privilige all time, while Aurora was running as Admin. I debbuged the output, and there was a Win32Exception all the time rust was in the foreground. I can post the logs later, but the documentation said windows ist throwing this exception when a 32 bit applikation

r3sist3nt:
I attached the logs of my and your branch. I added two logging lines to the original code in the `GetActiveWindowsProcessname()` to fetch the exception and the detected application path.
Have a look at line 86-101, that was the time Rust was open. It makes no difference if aurora is run as admin or not, the exception stays the same.
After my change, no more exception is thrown and RustClient.exe is recognized without any problems, see line 86-89 in debug_newcode.txt
Atleast when both processes run with the same privilige.
The Aurora-user, Rust-admin case does still not work here, but thats a generell privilige problem i think


[debug_oldcode.txt](https://github.com/antonpup/Aurora/files/752933/debug_oldcode.txt)
[debug_newcode.txt](https://github.com/antonpup/Aurora/files/752934/debug_newcode.txt)

r3sist3nt:
I found another bug. I fixed it, but i will test it for some days to be sure everythink is working fine now

r3sist3nt:
I am testing this detection method now for days and there was no false detection of any game i played.

antonpup:
It's not that I don't want to merge it. I was just wondering if it holds any improvements over the current method.

I'll merge it sometime later, for now, education takes priority for me.

Issue #523, opened 2017-04-09
Logitech G910(v2) nordic edition, wrong layout for localized keys


kaporten:
Using LGS 8.91.48 + Aurora v0.6.0b

Using an "Interactive Layer" with "Key Fade", I get the wrong keys lit up when pressing special-keys like 'æ', 'ø', 'å' or '-'. Basically any key to the right of the 0/P/L/. vertical line.

I've tried using all available layouts and none are correct. 

Is it possible for me to add a custom layout myself? If so, I'd be more than happy to experiment until I got the right layout and share the "Nordic" layout with you guys.

PR #525, opened 2017-04-11
Nordic localization for Logitech G910v2

kaporten:
Related to https://github.com/antonpup/Aurora/issues/523

I've added required "Nordic" localization files (Should cover Denmark, Sweden and Norway), and verified that all keys now light up as clicked using a simple Interactive Layer with Key Fade events on the Logitech G910v2.

**Remaining issues**
Only troublesome keys left are RETURN/ENTER which always lights up on the primary board, even if clicked on the numpad, but I read somewhere that this was not something that could be fixed, due to both keys being reported as the same keycode value.

A similar issue affects "minus" on the Nordic layout -- the key just to the left of the RSHIFT button. This is a minus symbol on Nordic layouts, and has the same effect as clicking minus on the numpad.


antonpup:
Thank you for your contribution. At a first glance, it looks well implemented. But unfortunately, I am busy, and will only be able to slowly begin working on Aurora sometime in next two weeks. So, my apologies in advance for a delay.

kaporten:
No rush, just happy to contribute with localization.

By the way - I feel there is missing some similar localization in "ToLogitechBitmap.ToLogitechBitmap", but the code there wasn't in use by the single use-case (key highlight) I knew how to test.

If you let me know how to test out that class as well, I can do a similar update there.

magnus97:
Spotted a flaw that the - key right next to . does not activate the right key fix for this is to change line 247: return DeviceKeys.NUM_MINUS; to return DeviceKeys.FORWARD_SLASH;

kaporten:
You're right - the OemMinus key next to . did not highlight, but instead the minus on numpad did. I thought that was just due to same keycode being sent (as with the 2 Enter keys sending same keycode), but it wasn't. Thanks.

antonpup:
This is specifically for G910v2? Would this be functional with G910v1 or any other Logitech keyboard for that matter?

I would have to also make these changes for other supported brands.

kaporten:
No I don't think this is specific to any one keyboard -- the title was just from me discovering the issue on this board. After investigating and performing the code changes, I've not had to change any Logitech-specific files at all.

Only generic files such as KeyUtils and KeyboardLayoutManager.

I've only tested these changes on a Logitech G910v2 (only Aurora-compatible board I own), but I expect these changes to be appliccable to any Nordic-layout board.

antonpup:
I see that this PR is from master to master. Could you please redo it for dev branch, so I can include it in development.

kaporten:
Yes, sure - closing this PR

Issue #329, opened 2017-01-07
Support for Aura, NZXT, Prismtik


nyan257:
1. Asus Aura
2. NZXT Hue+
3. Prismatik for Lightpack
4. Steelseries Engine

brainbug89:
1. Asus Aura doesn't have an API/SDK
2. NZXT Hue+ doesn't have an API/SDK
3. Prismatik could be possible. I don't know if there is an API/SDK
4. Steelseries is already on the [list](https://github.com/antonpup/Aurora#planned-featuressupport)

For Asus Aura and NZXT Hue+ you need to ask the manufacturers for API/SDK support.

SpencerCornish:
@brainbug89 Correct me if I'm wrong, but I believe that AURA does now have a public SDK. Link: https://www.asus.com/campaign/aura/us/SDK.html

brainbug89:
You're right. There is a newer issue for Asus Aura: #546

simon-wh:
@arch-master Yes, these will be dealt with in time. I've currently got exams so I can't do much and I've got some big changes that need to take place before I try to work on these myself.

Gurjot95:
Both Aura and NZXT are WIP. Also closing because of duplicate.

tirrorex:
@Gurjot95 hello, any ETA on this ?
I might be able to help with the programming if you give me some pointers on what needs to be done
(Would need to refresh my c++ skills though as i mostly do swift those days)

PR #645, opened 2017-07-08
SteelSeries GameSense beta SDK  support and Apex M800 keyboard layout

### LINKED ISSUE #162: Steelseries
Hi I have a MSI GE72 Laptop. I would love to see Steelseries support.
sdk is here:
https://github.com/SteelSeries/gamesense-sdk
Any chance of having your awesome software working here?

brainbug89:

Fixes #162 and number 4 of #329

- support for SteelSeries devices via GameSenseSDK
- first time install instruction for SteelSeries GameSense
- added SteelSeries Apex M800 keyboard layout

**Known issues/To do:**

- to do: test with Apex M800 keyboard


simon-wh:
@brainbug89 Thank you for this contribution! I tested it with my Rival 300 CS:GO and everything worked great! I'll try and get Mice images setup before the next release and hopefully, we can get someone with the Apex M800 to make sure everything is in order.

Project-Aurora/Project-Aurora/Devices/SteelSeries/GameSenseSDK.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private void setupLISPHandlers()
+        {
+            String json = "";
+            GameSensePayloadLISPHandlerJSON payload = new GameSensePayloadLISPHandlerJSON();
+            payload.game = sseGameName;
+
+            // sending POST requests with golisp handler
+            payload.golisp = "(handler \"PERIPHERYCOLOR\"  (lambda (data)    (let* ((val (value: data)))      (on-device \"rgb-2-zone\" show: val) (on-device \"rgb-1-zone\" show: val))))";
```
ToadKing: The GameSense API expects all the GoLisp handlers to be in one file, which this call creates. Because you call them every time they launch, the left-over handlers are still used from the previous calls so it still works but putting all these handles into one string and one call to `load_golisp_handlers` is the correct way to do it.

ToadKing:
Hi! SteelSeries dev here. I added a quick comment about the way you load the GoLisp handlers and have another quick comment. For handlers created in GoLisp, you should define the zones they write to using [`add-event-per-key-zone-use` and `add-event-zone-use-with-specifier`](https://github.com/SteelSeries/gamesense-sdk/blob/master/doc/api/writing-handlers-in-golisp.md#device-initialization-and-deinitialization) so the SteelSeries Engine frontend knows what zones/keys these events handle. For the events you use here the "all" zone should be good enough for all the handlers.

Project-Aurora/Project-Aurora/Devices/SteelSeries/GameSenseSDK.cs
brainbug89: That makes sense. :smile: Thanks for helping.

PR #647, opened 2017-07-09
Support for Borderlands 2

LukasACH:
- Adding Game support for Borderlands 2

**Known issues/To do:**
There are several points that dont work yet:
- ~~I can't get the profile to show up in the vertical profile selection bar, however the profile activates upon launching the game.~~
- The JSON-reader returns null instead of the content of the file. (The addresses are currently hardcoded)
- ~~Both PercentageEffectLayers don't show the correct values.~~

In some cases, I could't figure out the meaning of some variables as they only come up once?!

The profile is a mix between the RocketLeague and the CSGO profile.
I appreciate any help, however, I'm unable to work on this PR until 7/19.


LukasACH:
**Planned effects**
I have implemented a bar for Health and Shields as I could't find any other pointers witch yield the same value every time, eg. ammunition (I only found current, not the clip size), grenades and played character.

I could think of special effects if the shield breaks or if the character gets hurts (if current health is lower than the previous value)

I will start working on this when the current state is working.

mmalyska:
In LightingStateManager.cs in Initialize method add your application.

LukasACH:
I have added the [line](https://github.com/antonpup/Aurora/pull/647/files#diff-3cfd505ae5f3b4232b38214886f35f21R105) `new Borderlands2.Borderlands2()` in there. Where else do I need to add some code?

mmalyska:
Change yours png's "Build Action" from Content to Resource.
https://github.com/antonpup/Aurora/pull/647/files#diff-1c5aeaf896781f432c3e5f4d3e31ee6f
https://github.com/antonpup/Aurora/pull/647/files#diff-7988fdd956bafa60b231387a18a71958

LukasACH:
The Broderlands 2 Layer now works as intended. The Health and Shieldbars are now integrated into this Layer as there is no possibility to deactivate a separate Layer based on Logic.

simon-wh:
Looks good, Thanks for the contribution! I'll get BL2 downloaded to see if I can get other game information

Issue #665, opened 2017-07-20
Equalizer scaling with windows sound setting


Zoidster:
Hi and thanks for making this great software!
My suggestion is to add scaling to the equalizer so it still shows correct colors even when the windows volume is very low.
My use case is that I plug in my headphones into my speakers and lower the windwos volume quite a bit so my ears aren't blasted off.
I hope I could make clear what I mean :D

PR #667, opened 2017-07-21
Added scaling with System sound as a checkbox to the Equalizer settings

Zoidster:

Implements #665 

I suggested it and then I thought I'd just try to implement it, so here goes nothing.
Its just an additional checkbox in the equalizer settings and the backing property, also an additional variable in the render method of the equalizer.


simon-wh:
Looks good, thanks for the contribution!

PR #669, opened 2017-07-21
Update SteelSeries GameSense support

ToadKing:
This builds upon #645.

* Fixes malformed JSON in the keyboard payload. This fixes M800 support.
* Changes to GameSense handlers. Instead of different handlers for each type of event, use one handler for all color events. Works better with the SteelSeries Engine frontend.

Improves on #162.


simon-wh:
Thank you very much for helping out! I was going to have a look at what you mentioned in #645 later, thanks for saving me some time!

brainbug89:
Your GoLisp code looks way better than mine. Thanks again for helping out. :+1:

Vjgizmo:
Thank you @ToadKing 
I tested with my mouse Steelseries rival 100, with the generic mouse profile, it works very well.
So I created the rival 100 mouse image and modify generic_peripheral.json, I can share it for future integration.
[Image mouse + keyboard](http://imgur.com/a/8UOxd)

Issue #623, opened 2017-06-26
Option to Keep Key Illuminated When Held Down [Request]


User5432112345:
When using the "key fade" mode on an  interactive layer, the keys begin fading to the secondary color immediately after being pressed even if they are held down.  It would be good to have the option to keep them on the primary color for as long as the key is held down, then to begin fading on release.  I think this should be the default behavior, but having the option would be even better.

simon-wh:
This was added thanks to PR #676 and is available in development releases.

PR #676, opened 2017-07-22
Added Interactive Effect option to start progress of effect when key is released

DogeAmazed:
Implements #623 

Hope everything is okay with the code style.


simon-wh:

PR #688, opened 2017-07-26
Fixed broken Numpad-Keys when Numlock not active

### LINKED ISSUE #444: Wrong keys light up. (Logitech g910, swedish/nordic layout)
I've discovered a bug on the g910 which leads to wrong keys being lit up when using an interactive layer.

I have a Swedish layout Logitech g910 Orion Spectrum.
It only happens when num lock is off and only the numpad keys are affected.

When num lock is off the following keys light up instead of the normal ones.

- 1 - End
- 2 - Down arrow
- 3 - Page down
- 4 - Left arrow
- 5 - (Nothing lights up)
- 6 - Right arrow
- 7 - Home
- 8 -Up
- 9 - Page up
- 0 - Insert
- Comma - Delete

(This one is always incorrect) - Enter - Lights up the enter key below backspace instead of the numpad one.

I know this is what the buttons do when num lock is off and I can understand why it happens.
If I disable Aurora and use logitechs own effect it always lights the correct keys.

DogeAmazed:
- Fixed #444 

N/A

**Known issues/To do:**
N/A

Well, I don't know why the normal keys are the extended ones...

### DIFF
--- a/Project-Aurora/Project-Aurora/Utils/KeyUtils.cs
+++ b/Project-Aurora/Project-Aurora/Utils/KeyUtils.cs
@@ -25,6 +25,8 @@ namespace Aurora.Utils
             {
                 case (Keys.Escape):
                     return DeviceKeys.ESC;
+                case (Keys.Clear):
+                    return DeviceKeys.NUM_FIVE;
                 case (Keys.Back):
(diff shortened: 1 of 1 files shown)


simon-wh:
Awesome! Thank you very much for the fixes!

PR #689, opened 2017-07-26
Fixed animator crash

### LINKED ISSUE #684: Animation editor crash
When editing any of the values in the animation editor, if I backspace the last character present it crashes. If I instead select the text with mouse cursor and write over it, it doesn't crash.

Using version 0.6.1-dev.

DogeAmazed:
- Fixed #684 

N/A

**Known issues/To do:**
N/A

You are to fast for me, otherwise this would be in the other PR 🐕


simon-wh:
Looks good! Thanks!

Issue #145, opened 2016-09-15
Pressing Alt Gr illuminates keys of Ctrl shortcuts


brainbug89:
Pressing the Alt Gr (right Alt) key illuminates keys of the Ctrl shortcuts:
![aurora_altgr_shortcut_bug](https://cloud.githubusercontent.com/assets/21130949/18555116/e68764fe-7b66-11e6-8e8f-e42506c9352f.PNG)

Keyboard: K70 RGB Rapidfire (DE layout)

simon-wh:
This is no longer an issue due to the Shortcut Assistant redesign and changes made in PR #693

PR #693, opened 2017-07-27
[Interactive Layer] Fixed LControl lighting up, when Alt GR is pressed

### LINKED ISSUE #311: Interactive layer: Alt Gr and Ctrl flickering for key fade
Using Key Fade and disabled Shortcut layer: the **Alt Gr** and **left Ctrl** key start flickering while pressing **Alt Gr**.

Could be related to #145

_Version: v0.6.0
Keyboard: Corsair K70 RGB, German_

DogeAmazed:
- Fixed #311

N/A

**Known issues/To do:**
- Somebody should look also into #145 (Same issue , but in ShortcutAssistantLayer), if someone does it before me, this PR can be accepted. If not, I include the fix into this.

### DIFF
--- a/Project-Aurora/Project-Aurora/InputEventsSubscriptions.cs
+++ b/Project-Aurora/Project-Aurora/InputEventsSubscriptions.cs
@@ -50,11 +50,23 @@ namespace Aurora
 
         private void Internal_KeyUp(object sender, KeyEventArgs e)
         {
+            KeyEventArgsExt args = (KeyEventArgsExt)e;
+
+            //This is so we can ignore the LControlKey when Alt Gr is pressed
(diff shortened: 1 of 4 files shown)


simon-wh:
Looks good! Thanks for the fix! With regards to #145 because of the changes to the shortcut assistant layer it means that the problem with it lighting ctrl shortcuts isn't a problem as you assign a specific set of keys, but you still can't assign Alt Gr as a shortcut as it will result in both LCtrl and RMenu being in the keybind. To fix this I've added a check to the Internal_KeyUp and Internal_KeyDown to ignore the event if the KeyUtils.GetDeviceKey call returns DeviceKeys.None which allows you to use Alt Gr with the shortcut assistant without an issue. However, I think I should probably add some form of option to the Shortcut Assistant layer so that it will trigger LControlKey presets when LControl or RControl is pressed.

Issue #601, opened 2017-06-13
Peripherals stop working while Aurora is active [Overwatch]


TekkertheChaot:
### What are you experiencing an issue with?

- [x] Aurora

### What kind of an issue is it?

- [x] Performance problems

### Actual behavior:
When I play Overwatch with Aurora activated my input lag increases and often my peripherals stop working for up to 3 seconds (including input and lighting effects)

### How often does this problem occur?
Everytime when I play Overwatch with Aurora active in the background

### What is the version of Aurora you are experiencing this issue with?

My version is v0.6.0b

I wasn't able to find a to my problem related ticket. 
Hope this wil help.

simon-wh:
What devices do you have? Also, could you try the [latest development release](https://github.com/antonpup/Aurora/releases) to see if it is any different on that build?

simon-wh:
Any updates on this?

TekkertheChaot:
I'm sorry Simon, I haven't got any notifications about this ticket until now.
So, sorry for leaving you hanging there 2 months

To answer your comment from about a month ago: I use the Logitech G810 Orion Spectrum (keyboard) and the Logitech G502 Proteus Spectrum (mouse). 
Since posting here, I continued playing with aurora active in the background and the issue seem's to be gone after a major Windows update (something after the Creators update). The issue still appears in heavy fights (much effects on screen) even though my CPU performance isn't peeking out to its maximum.

I will definitely try the development release and report back end of this week to stress-test it out.

simon-wh:
@TekkertheChaot Thanks for getting back to me. Currently, @VoronFX is working on changes which will hopefully eliminate most if not all input lag issues which can be seen over here #561.

PR #708, opened 2017-08-02
Redesign of input classes to use Raw Input

### LINKED ISSUE #561: Mouse Input Lag/Visible Effects Lag after keyboard change
### What are you experiencing an issue with?

- [x] Aurora
- [ ] Aurora supported game, (Fill in the game)
- [ ] Aurora Wrapper
- [ ] Aurora Profile
- [ ] Aurora supported device, (Fill in problematic devices)
- [ ] Aurora Scripts
- [ ] Aurora Updater
- [ ] Aurora Lighting issue
- [ ] Other, (Fill in another issue here)

### What kind of an issue is it?

- [ ] Crash
- [x] Performance problems
- [ ] Visual bug
- [ ] Localization issue
- [ ] Enhancement suggestion
- [x] Device issue
- [ ] Other, (Fill in another type of issue here)

### Expected behavior:
Lighting to work, no input lag.

### Actual behavior:
What is happening is when I updated my keyboard setup (bought a Razer Ornata) and swapped keyboards, I am getting considerable input lag spikes with my G303 Mouse and the lightning effects also lag as they run on the keyboard. I noticed the lighting effect chop up greatly when achieving an ultimate on Overwatch. All Driver versions for my keyboard and mouse are to the latest spec but the spikes seem to still happen.

This problem occurs... Not just in games but in normal desktop use too. Aiming in FPS and first person perspective games gets chopped up every so often. Additionally, there is some CPU usage spikes from the program itself that happen rarely.

N/A

My version is... v0.6.0b

The previous version, both this and previous versions of Aurora did not do this before to where the input lag is noticeable.

VoronFX:
**List any issues that this PR fixes:** fixes #561, #601

- ```InputEventsSubscriptions``` renamed to ```InputEvents```, moved to Utils\InputEvents.cs  and redesigned to use Raw Input for capturing.
- New class ```InputInterceptor``` under Utils\InputInterceptor.cs. It uses lightened version of Low Level Hook and dedicated high-priority thread to decrease a chance of delaying input in system. Some code and classes were dublicated from MouseKeyHook as they are internal.
- ```InputInterceptor``` is used only if there is necessity for it (when "volume as brightness" enabled).
- SharpDX.RawInput NuGet package added as dependency. Contains useful classes for using Raw Input.
- Found and fixed: ```KeyUtils.GetStandardKey``` returned wrong value for ```Keys.RWin```.
- Other small fixes according to redesigned input.

**Known issues/To do:**

- Low Level Hooks doesn't works while Raw Input is used and application has focus. As workaround Raw Input can be disabled while application has focus. But I consider this as not so important to write a lot of workaround code. Instead I added notice to "volume as brightness" (the only feature that is affected) that it doesn't work while Aurora in focus.

 **Related articles and information:**
As there is very few info about window input in internet, I am going to leave here links to useful articles.
- [Nice article about messy input API's and how to handle Raw Input.](https://blog.molecular-matters.com/2011/09/05/properly-handling-keyboard-input/)
- [Using Raw Input from C# to handle multiple keyboards.](https://www.codeproject.com/Articles/17123/Using-Raw-Input-from-C-to-handle-multiple-keyboard)
- [C# lib for HID using Raw Input. Another one implementation.](https://github.com/Slion/SharpLibHid)
- [Combining Raw Input and keyboard Hook to selectively block input from multiple keyboards](https://www.codeproject.com/Articles/716591/Combining-Raw-Input-and-keyboard-Hook-to-selective)


simon-wh:

PR #729, opened 2017-08-11
Updating boost pointers for RL

th3an7:
- Update for Rocket League boost pointers

**Known issues/To do:**
- Team/score pointers are still broken
- Test for RGB keyboard users


simon-wh:
Thanks for the contribution! I'm just going to merge this without testing as it is currently broken and if it doesn't work properly then I'll just sort it out when I come to sorting out the rest of the Rocket League stuff for the next public release.

PR #732, opened 2017-08-13
Small fixes in ColorSpectrum and PercentEffect.

VoronFX:

- Fix: ```ColorSpectrum.ToLinearGradient``` ignored shift value.
- Fix typo in ```ColorSpectrum.Shift```.
- Fix: piece of last gradient color in PercentEffect sometimes was also drawn in the beginning due to float error.

### DIFF
--- a/Project-Aurora/Project-Aurora/EffectsEngine/ColorSpectrum.cs
+++ b/Project-Aurora/Project-Aurora/EffectsEngine/ColorSpectrum.cs
@@ -6,310 +6,321 @@ using System.Linq;
 
 namespace Aurora.EffectsEngine
 {
-    /// <summary>
-    /// A class that represents a spectrum of colors. After creating an instance, you can then retrieve a blended color within a range of [0.0f, 1.0f].
-    /// </summary>
-    public class ColorSpectrum
-    {
-        /// <summary>
-        /// A predefined rainbow ColorSpectrum
-        /// </summary>
-        public static ColorSpectrum Rainbow = new ColorSpectrum(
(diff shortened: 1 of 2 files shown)


VoronFX:
Arrrrgh. These line endings... =_=
I've just counted that project has only 0.0773% of linux style ("\n") line endings. But these 
0.0773% are almost in any file, causing annoying messages about inconsistent line endings and git comparing changes wrong. Maybe we should normalize all of them in repository and/or tune .gitattributes?

simon-wh:
I've added a gitattributes file which will hopefully sort out that issue. Also, thanks for these changes!

brainbug89:
What kind of line endings are we using now? I'm used to linux style, but I can change my settings if necessary.

VoronFX:
Last time I counted that repository mostly has crlf line endings (92.27%). But ```.giattributes``` now has ```* text=auto``` setting, which means that all committing files will be converted to lf line endings if git decides that file contains text. So in future there should no crlf endings left.
Also we should be aware that git can corrupt files if he suddenly decides that some binary file with extension that git doesn't know contains text.
[Here](https://stackoverflow.com/a/38017715) is recommendation to use explicit setting for file extension instead of for all files. 
Decision on which approach should we use I am leaving to @simon-wh .

simon-wh:
@VoronFX I've removed the auto line and I've changed it so that .cs is set to crlf. If most of it is currently crlf then it would make sense to stick with that. Unless there is a significant reason to switch to using unix fully

PR #738, opened 2017-08-15
Getting process names fixes

VoronFX:

- Some apps and games (like Blade & Soul) runs with admin rights (i.e. evevated). Built in ```Process.MainModule.FileName``` uses old function that throws access denied exception when target process is elevated, despite of calling process is elevated or not. 
In Vista and above exists a new win function for this task. I've changed code to use it. 
[Here is an article about this bug.](http://www.aboutmycode.com/net-framework/how-to-get-elevated-process-path-in-net/)
- Fix inconsistent lower-casing when comparing process names. In ```LightingStateManager``` ```RegisterEvent()``` used ```ToLower()``` while comparing in ```Update()``` was using ```ToLowerInvariant()```.


simon-wh:
Awesome! Thanks for the contribution!

PR #748, opened 2017-08-18
Add note about League of Legends EnableLightFx config

lucasvanhalst:
For the League of Legends effects to work, you need to set `EnableLightFx` to `1` in the game.cfg file. This currently wasn't described anywhere except in a couple of issues in this repo.


simon-wh:
Thanks for the contribution!

PR #759, opened 2017-08-23
Fix G keys mapping for Corsair K95 Platinum keyboard

davidarellano:
- Small fix in the json for K95 Platinum, some "G keys" at the left were no lighting up at all.


simon-wh:
Thanks for the contribution!

Nevoska:
It still dont lightup when playing Overwatch for exemple. The light bar does not light up neither

PR #787, opened 2017-09-11
Added 'Write log file' option

kirbykirby56:
I did this because Aurora generated log files over 12 GB in size.

Thanks, and thanks for such a cool program!

### DIFF
--- a/Project-Aurora/Project-Aurora/ConfigUI.xaml.cs
+++ b/Project-Aurora/Project-Aurora/ConfigUI.xaml.cs
@@ -741,7 +741,7 @@ namespace Aurora
 
         public void ShowWindow()
         {
-            Global.logger.LogLine("Show Window called");
+            if (Global.writeLogFile) Global.logger.LogLine("Show Window called");
             this.Visibility = Visibility.Visible;
             this.WindowStyle = WindowStyle.SingleBorderWindow;
             this.ShowInTaskbar = true;
--- a/Project-Aurora/Project-Aurora/Controls/ArrangeListBox.cs
+++ b/Project-Aurora/Project-Aurora/Controls/ArrangeListBox.cs
@@ -145,7 +145,7 @@ namespace Aurora.Controls
 
                 index = Math.Min(index, Children.Count - 1);
             }
-            //Global.logger.LogLine("Setting order to " + index);
+            //if (Global.writeLogFile) Global.logger.LogLine("Setting order to " + index);
             bool order_Changed;
             if (order_Changed = index != GetOrder(_dragging))
                 SetOrder(_dragging, index);
@@ -201,7 +201,7 @@ namespace Aurora.Controls
             Rect position = GetArrangement(d);
             if (double.IsNaN(position.X))
             {
-                //Global.logger.LogLine("No anim for me");
+                //if (Global.writeLogFile) Global.logger.LogLine("No anim for me");
                 SetArrangement(d, desiredPosition);
                 InvalidateArrange();
                 return;
@@ -217,7 +217,7 @@ namespace Aurora.Controls
 
         private void ReorderOthers()
         {
-            //Global.logger.LogLine("Reorder others");
+            //if (Global.writeLogFile) Global.logger.LogLine("Reorder others");
             int s = GetOrder(_dragging);
             int i = 0;
             foreach (var child in Children.OfType<UIElement>().OrderBy(GetOrder))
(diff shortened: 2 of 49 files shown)


simon-wh:
If Aurora is producing large log files the solution is to stop it from logging unnecessarily (as it isn't designed to do that), not to stop it logging altogether. Logging is crucial for a lot of debugging. Could you provide a sample of the large log files you're getting?

kirbykirby56:
Unfortunately, I deleted them permanently. 

Writing logs is on by default, but it's just a way to stop them for people like me who don't have a problem with the app, other than its log files. I understand what you're saying though. I'll turn it on on my version and see what I get, then get back to you.

kirbykirby56:
Oh wait! I do have a log file at the ready, although it isn't huge. I'll see if I can upload it somewhere.

simon-wh:
@kirbykirby56 Thanks, A few things to keep in mind for contributing in the future is to make sure you are doing it on the correct branch, because in this case, it should be on dev, not master. Also, for the changes you actually made, it would have been easier to just change the start of the LogLine method rather than changing every instance of a call to the LogLine method. Especially considering there is a good chance that logging that would be added after that change would forget to include the if statement.

kirbykirby56:
I was trying to find the definition of LogLine, but had a bit of trouble finding it. 
I'll also keep note of the dev branch, sorry about that!

simon-wh:
If you're using Visual Studio you should be able to press F12 and it'll take you to the definition of the method.

kirbykirby56:
It kept saying "Could not find the symbol under the carat."

simon-wh:
Strange, I don't have that issue, but I believe I've had Visual Studio installs before where it didn't work properly. The logger is under Utils/Logger.cs, but it is no longer used in the dev branch as I've switched over to using NLog.

kirbykirby56:
When I start a modification, should I also check out from dev branch?

simon-wh:
Yes, as that means you'll be changing to most recent development code base.

kirbykirby56:
Alright, cool.
The main thing that seems to be logging is as I sent above.
I think it might be because those devices are enabled in the Device Manager in Settings.
Maybe a fix would be to have those on "disabled" until Aurora detects that device?

simon-wh:
@kirbykirby56 I plan to add checks so that Aurora can actually detect if the devices are available so it doesn't attempt to initialize them unnecessarily. You can just disable the devices that aren't relevant to you in the device manager for now.

kirbykirby56:
Yeah, I did that. I'll get back to you if I get large logs again! Thanks for the pointers, too. :)

simon-wh:
Cool, no problem!

kirbykirby56:
Hey! I'm getting large files again. My Aurora logs folder is now 5.8 GB! I decided to pick out a particularly large one (1.57GiB) and zip it for easy sending. It zipped to around 6 MiB, or .3% of the file size, which means there's a lot of repeated info. I've uploaded this file to Mega (at [Mega.NZ](https://mega.nz/#!hnhVnJaD!QQwI6q8m96HNM2ZUSvXEeJ7EawbO_GbWhQkCHcr32fM)), but the file may be too large to open. As such, I also compressed a much smaller file, (~500 MB) as well. This time, it's that the Logitech LED SDK couldn't initialize. It logs around 20 times a second, with around 4kb/s. The main issue is that the program never stops logging this information, and was logging for three days. I can't even open larger files to see if they are logging the same thing. Here's the smaller log. [Mega.NZ](https://mega.nz/#!5np1FQYL!TpTOpFJWtcyGDE4cZRomX7bv-gRz74Iyl_6QVwVBA18 (1.31MiB).

A good way to stop this from happening is check once every, say, five seconds, or don't log unless the logs are different. (Maybe queue for five seconds, and if the message is the same as the last successful log, stop it from writing). If one line was created every five seconds, instead of twenty times each second, the log file would be 100x smaller than it currently outputs for me. It's also possible that the logs could be much larger for other people because I have the other types of devices disabled in Device Manager.

Thanks!

kirbykirby56:
If it helps, this seems to be the offending line. 

> [ERROR] 10/28/2017 8:25:36 PM: Logitech LED SDK could not be initialized. (LogiLedSetTargetDevice or LogiLedSaveCurrentLighting failed)
> [INFO] 10/28/2017 8:25:36 PM: Device, Logitech, was not initialized
>

simon-wh:
I'm not sure why that is happening as it should stop attempting to initialize it after a few retries. Do you have a logitech device? Because if you don't you should disable it as was mentioned above.

kirbykirby56:
Yeah, I have a Logitech device. It allows Aurora to handle lighting, too.




On Sunday, November 12, 2017, 11:08 AM, Simon W <notifications@github.com> wrote:


I'm not sure why that is happening as it should stop attempting to initialize it after a few retries. Do you have a logitech device? Because if you don't you should disable it as was mentioned above.

PR #857, opened 2017-11-07
Razor macro key naming fixed

DogeAmazed:
See [here](https://www.razerzone.com/gaming-keyboards-keypads/refurbished-razer-blackwidow-chroma-v1).


simon-wh:
Thanks for the contribution!

PR #1102, opened 2018-05-18
Updates to packages and experimental SteelSeries HID Support 

SnakePin:

- Updates NuGet packages.
- Changes platform toolset and windows sdk to a newer version to ensure compatiblity with latest visual studio.
- Adds experimental steelseries hid write support.

**Known issues/To do:**

- To do: Use mono usb library instead and add other rival mouses support.
- 
-


simon-wh:
Thanks for the contribution! I've had a bit of a look around the code and I don't really like the hardcoded hex values that are used for communicating with the device, it may make sense to create a base class for devices and create sub-classes for each device which deals with the particular interactions?

SnakePin:
What do you mean by sub-classes, different class for each rival mouse? Wouldn't that create mess?

SnakePin:
Damn this close and comment button is big.

simon-wh:
What would create a mess would be having if statements in every method that interacts with a device, for every device. It would make a whole lot more sense to just have one class which handles the interaction for each device (as for stuff like Rival 300, and 300 CSGO all you need to change is the identifiers) and just swap out which class is being used based on what devices are found.

Having the main class not be changed for every device you add is very powerful. If all you need to do is create a new sub-class for, let's say the rival 700, you change any particular interactions that are different and that's it, that makes everything a whole lot nicer. Things will get really messy if you start hardcoding interactions and putting if statements in places where there are interactions, it makes it more awkward to find & fix bugs and more awkward to add new devices.

SnakePin:
Oh ok I get it. I'm on it now.

SnakePin:
Is this what you asked dude? I added new commit, it has better performance and I added other devices but Rival 310 and Rival 700 is not reverse engineered yet so I couldn't add their support now. For Rival 500 It looks complicated. And sorry but the HidReport can only be local variable.

SnakePin:
I made it modular, other rival mouse supports can now be added easily.

SnakePin:
Sorry for inconvenience boss, now It is same as you wanted.

simon-wh:
@SnakePin It's looking a lot better! There are some more changes I want to make, but I'll deal with those, you've already done a lot of really good work! Let me know whenever you've finished making your changes for this PR, then (when I'm free after exams) I'll make some finishing touches and merge! Thanks again!

SnakePin:
I think I my job on this PR finished for now. But what changes do you want to make, maybe I can do that.

simon-wh:
Mainly just cleaning up the code related to the classes. i.e. create a list of the different device classes that is iterated through for checking them all, a list of the ones that have been found instead of having individual bools and simplify some things. You don't need to worry about it, I don't mind doing it, but you can try and do some of it if you want.

PR #1106, opened 2018-05-24
Simple Conditional Layer

Wibble199:
Adds a simple conditional layer which allows for key colors to be set based on a boolean value read from a game state integration.

CS:GO Conditional Layer Example


simon-wh:
Everything looks good! Thanks for the contribution!

PR #1114, opened 2018-05-30
Correct issues with Razer Blade keyboard mapping

### LINKED ISSUE #776: Razer Blade (Non-Stealth) Keyboard "Offset"
### What are you experiencing an issue with?

- [ ] Aurora
- [ ] Aurora supported game, (Fill in the game)
- [ ] Aurora Wrapper
- [ ] Aurora Profile
- [Razer Blade 2017 FHD (Non-Stealth)] Aurora supported device, (Fill in problematic devices)
- [ ] Aurora Scripts
- [ ] Aurora Updater
- [ ] Aurora Lighting issue
- [ ] Other, (Fill in another issue here)

### What kind of an issue is it?

- [ ] Crash
- [ ] Performance problems
- [ ] Visual bug
- [ ] Localization issue
- [ ] Enhancement suggestion
- [x] Device issue
- [ ] Other, (Fill in another type of issue here)

### Expected behavior:
Keys F1 to F12 should light up
Keys 1 to 6 should light up

### Actual behavior:
Keys ESC to F11 light up
Keys ~ to 5 light up

This problem occurs for all profiles for the Razer Blade. It's a keyboard layout mapping issue (I think).

1. Select "Razer = Blade" while using a Razer Blade (non-stealth).
2. Select a default profile.
3. Look at the keys.

My version is **beta v0.6.1-dev3**

The previous version did not (directly) have this issue. Since there was no Blade profile, I made use of the Razer BlackWidow Tournament Edition keyboard layout since it was almost exactly the same as the one on the Blade. However, it seems to have gained an "offset" too in the mapping in this update.

### Please include any logs, screenshots, or files that are related to this issue:

For logs, you can use: http://pastebin.com/
For screenshots, you can use: http://imgur.com/
For files, you can use: http://www.tinyupload.com/ , https://www.dropbox.com/ , or attach them to the issue.

Pictures on request.

NoOne-dev:

Fixes #776

- Keyboard remapping on the Razer Blade

**Known issues/To do:**

- Possible issues with new ARROWDOWN mapping
- Issues with FN left and right mapping
- Possible issues with left and right CTRL, ALT, WIN key mapping


simon-wh:
Thanks for the contribution! Are you planning to fix those other things you mentioned in this PR or shall I merge this?

NoOne-dev:
I was wondering if you had any insight into the odd interactions with the up and left arrow keys. I can't seem to track down what causes them to combine when being Chroma-mapped. I also for some reason wasn't able to get the "\" (Backslash) key to work correctly despite the actual mapping being correct. It seems like the UI element to show that key is missing/not working too.

simon-wh:
@NoOne-dev I pushed a commit that fixes the loading of the layout, it was erroring out in the middle of adding the 'extra features' so the backslash and any mouse that was selected wasn't being added. However, I'm not sure what you mean by "the odd interactions with the up and left arrow keys"?

NoOne-dev:
The up and left arrow keys, which as far as I can tell from the Chroma SDK documentation don't have any special interactions, seem to share the same ID for Chroma LED mapping.

simon-wh:
@NoOne-dev Okay, is that everything sorted for the layout? (assuming you're using a Razer Blade)

NoOne-dev:
Down Arrow, Right Arrow, Space, and Backslash are all not mapped to the correct values still. Enabling a color or effect on the Up Arrow causes both the Up Arrow and Right Arrow to light up. So far haven't been able to come up with a solution.

NoOne-dev:
It's not perfect yet, but it's much better then what's currently in place. Please merge this.

NoOne-dev:
After trying the current version, it seems some of my key maps were broken at some point after this because the secondary keys such as Tab, Caps, and Shift are now incorrectly mapped.

NoOne-dev:
It works now, and it's spectacular.

PR #1136, opened 2018-06-17
American Truck Simulator support

Wibble199:
Added American Truck Simulator support (uses the same GSI/Telemetry DLL as Euro Truck Simulator 2).


simon-wh:
Thanks for the contribution!

PR #1138, opened 2018-06-17
Improved Default Dying Light Profile

diogotr7:
This should cover most if not all the usable keys in-game.


simon-wh:
Thanks for the contribution!

Issue #1017, opened 2018-03-08
Key Color Toggle


RichHammond:
Hi!
I'm loving the software right now, it works wonders on my keyboard.

I'm wondering if it would be possible for you to add a toggle on keypress.
I want my pause key to be one color to start, then switch to another color once i press it, and the next time I press it switch back to the first color.

This would be useful for recording toggles like OBS so I can have one color when I'm recording, and one color when I'm not.

diogotr7:
I'd like this too. I currently have my scroll lock key set like this and bound to mute on discord, but it would be useful if it were more flexible.

piercy:
im not sure if its in the released build, but there is a lock color layer in the dev build im using, so might be available in the beta / alpha / nightly builds.  It does only support capital, scroll lock and numlock atm.  But maybe using scroll lock would be acceptable for you? 

![image](https://user-images.githubusercontent.com/1812670/37179333-3355af52-231d-11e8-99b5-d35fea329516.png)

RichHammond:
I'm on beta 0.6.0 and have that, but I don't want it with the lock keys, I want it with pause.

simon-wh:
Added thanks to @Wibble199 with #1139

PR #1139, opened 2018-06-18
Added a Key Toggle Layer

Wibble199:
Added a key toggle layer (based on the suggestion from #1017) that toggles a sequence on/off when one or more key combinations are pressed.

UI

In the above example, when Ctrl + F8 are pressed, the red keys change to green. When Ctrl + F8 is pressed again, they change back to red.


simon-wh:
Everything looks good! Thanks for the contribution!

murphys1987:
This is getting me closer to what I'm looking to accomplish, can you think of a way to configure this so that any key plus the entered trigger key enables the toggle? 

I'd like Caps Lock, C, and or R to always toggle this layer, however if i'm holding any other keys down it won't toggle the effect.

I know I could certainly add in W+Caps Lock, W+R, W+C in along with all the other permutations of it, however, that may be come a nightmare.

Maybe an "Any Key(S)" + R option would be possible?

Issue #1077, opened 2018-04-27
Fast switching profiles


Deskowicz:
Is it possible to fast switch between Profiles? I want to make 2 profiles to Fortnite, one with building keys backlight, and second with shooting keys.

I bind "Q" as a key to both profiles, but it works randomly, or really slow, because sometimes it switch immediately, and sometimes i need to press key 2-3 times or more to switch them.

Regards

simon-wh:
Hm, without looking at the code specifically I assume assigning both to the same key results in the first/last one it checks being the one loaded. I'll add a cycle/toggle option that'll allow you to swap between different profiles using a particular key.

simon-wh:
Added thanks to @Wibble199 with #1143

PR #1143, opened 2018-06-19
Improved Profile Keybind Switching

Wibble199:
Improved the logic for keybind switching so that it behaves more reliably when multiple profiles have the same keybind (#1077).

Currently it cycles through profiles in the order they were created, so some way of changing the order should probably be added in future though this works for now.


simon-wh:
That's a really good solution! Thanks for the contribution!

simon-wh:
I just tested it and with only one profile with the keybind it only switches when assigning a keybind, sorting it out now.

Wibble199:
Ah yes I see what I did. I had `if (Controls.Control_Keybind._ActiveKeybind != null) return;` and then moved it into that other if and forgot to swap the condition >.>

Thanks!

Issue #710, opened 2017-08-03
Add a mouse? - G502 (with images made)


Kisco:
I made this mouse image which is the g502, Link to the image is here - http://imgur.com/a/Va354

Note: If you would want anything fixed up just ask. No feelings if you already made one I just made this in my free time.

Edit: Updated link

Kisco:
@simon-wh Forgot to add the DPI lights! Will fix up tomorrow (3 am right now)

simon-wh:
@Kisco Okay, thanks.

Kisco:
Updated! @simon-wh - http://imgur.com/a/Va354

Kisco:
Fixed some edges, added dpi lighting and also corrected the outside color to be darker the the lines inside.

simon-wh:
Do you have photoshop files for them?

Kisco:
paint.net lol :V

Kisco:
i could transfer the layers to photoshop if you would like

simon-wh:
I've got paint.net installed, so you can throw over the paint.net files and I'll work with them

Kisco:
alright - https://mega.nz/#!iYYBVSIB!zE8Py0X2hYzogi-uMYjqCk_VjJuS4U5wx8enioIbZRw (with images and layers)

TomSputz:
Just curious - any progress with this?

simon-wh:
@TomSputz I haven't had time to get around to putting this in yet.

TomSputz:
@simon-wh Of course :) It's gotta be a lot of work put into this

I've just recently got the mouse, and have found out it actually has 2 lighting zones that aren't ordinarily controllable. I don't know if you have access to them through the logitech SDK, but a heads up

diogotr7:
@TomSputz the dpi lights are always the same color as the main G logo. To light them all up you have to create 2 dpi profiles lower than you currently use the mouse ( as long as you don't actually use the sniper button, of course)

TomSputz:
@diogotr7 Yes - that was the point I was making - you *can* actually control those individually - https://www.reddit.com/r/G502MasterRace/comments/6t0is6/how_to_overwrite_the_internal_g502_memory_with/ - Though I'm not sure the SDK can control that (I have tried to separate them, it works)

TomSputz:
@Kisco How did you make the pictures by the way? They're really nice

TomSputz:
[...]

PR #1146, opened 2018-06-21
G502

diogotr7:
Added support for the Logitech G502 mouse.
Solves #710 , thanks for the images @Kisco


simon-wh:
Thank you very much for the contribution!

PR #1150, opened 2018-06-23
Timed Key Layer

Wibble199:
Adds a layer whose sequence is lit in a single color. Then when a keybind is pressed (can be separate from the seqence) the sequence changes to another color for a set duration. After the duration, the keys return to their original color.

Can also be set so that the keys fade back to their original colour instead of abruptly changing.

Multiple "repeat actions" are available to customise what happens if one or more of the keybinds are pressed again while the timer is active:
- Reset (Default) - Restarts the timer when the key is pressed.
- Extend - Adds another duration onto the end of the existing one (so multiple presses stack).
- Ignore - Ignores a keypress while timer is active, and simply continues the existing one.

image

Thanks to _amahlaka97_ on Discord for the suggestion!


simon-wh:

PR #1151, opened 2018-06-24
Minor script layer improvements

Wibble199:
Cleaned up the interface for the script layer and added the ability to reload scripts.

Previously, Aurora would need to be restarted before script changes or new scripts would be recognised, which was a pain if you were actively developing and testing a script. With this change it is much quicker :)


simon-wh:
Cool! Thanks for the contribution!

PR #1152, opened 2018-06-25
Added mouse events to Global.InputEvents

Wibble199:
This allows for layers such as a layer whose highlighted keys depends on the scroll position (consider the toolbar at the bottom of games such as Factorio or Minecraft).

EDIT: Added a ToolbarLayer that utilised the mouse scroll event. This layer was as suggested by amahlaka97 on Discord and the scroll improvement was suggested by DrMeteor/diogotr7. When one of the keys in the sequence it uses is pressed, that key becomes "active". When another key is pressed, that key becomes active instead and resets the others to their default state. It's sort of like a radio button but for the keys on the keyboard. An option is provided that allows the mouse scroll wheel to be used to scroll up and down the active key.


simon-wh:
Thanks for the contribution!

PR #1167, opened 2018-07-10
Added Support for Quantum Conumdrum

diogotr7:
I could not get the game to work if i added a custom layer for it. I wanted to apply color enhancing, so I added a native layer.


simon-wh:
Thanks for the contribution!

PR #1168, opened 2018-07-12
Small typo in quantum conundrum

diogotr7:


gitmacer:
sure, you fixed it Quantum Conumdrum is still written as QuantumConumdrum.

diogotr7:
whoops, I fixed the wrapper patch being 32 bits instead of 64, didn't even notice that

PR #1183, opened 2018-07-18
Use desktop duplication for ambilight

lucasvanhalst:
Uses desktop duplication instead of taking screenshots for the ambilight layer. This is lighter on resources than the current implementation, and doesn't make WDM lag. Seems to be working well in games without too much of a performance hit.

I'm a novice at C# so the code could probably use improvements.

#### Known issues/To do:

~~Sometimes there's a System.AccessViolationException in this part after the desktop duplication has been restarted (e.g. when the config is changed):  https://github.com/lucasvanhalst/Aurora/blob/55e3f2923ea93bc84fba7a84814bfd466c7b0d03/Project-Aurora/Project-Aurora/Utils/DesktopDuplicator.cs#L102-L110 and I'm not sure why or how to fix it~~
Now fixed


diogotr7:
I just tested this and it's much faster than waht we currently have (30-40% cpu vs 5-10%...) It merges relatively cleanly, too. I'm not sure why this was abandoned but i think it's a great improvement.

lucasvanhalst:
Probably because of the issues I've mentioned. It crashes randomly and I don't really know why.

lucasvanhalst:
Hmm, just tested it for a few hours (with fullscreen applications as well) and it hasn't crashed yet. Could anyone else test this to confirm it's working properly?

diogotr7:
Will test, awesome work!

diogotr7:
I just linked this build on aurora's discord so hopefully ill be able to give some feedback.

PR #1213, opened 2018-08-05
Minecraft Support & Title-based Profile Detection

Wibble199:
## Minecraft Support

Added Minecraft support that uses a custom [Forge mod](https://gitlab.com/wibble199/aurora-gsi-minecraft) to send data to Aurora. Currently the mod only supports Forge with MC 1.12.2 (the most recent Minecraft version with Forge support), but I could probably add more if people want.

Data is sent via HTTP to the listener setup by Aurora. The values being sent to Aurora are:

- Whether the player is in-game (versus on the menu)
- Player health (color depending on whether the player is withered, poisoned, has the regen effect or not)
- Player absorption (golden hearts in-game)
- Player hunger
- Player armor
- Player experience
- Whether the player is burning
- Whether the player is in water
- Whether the player is sneaking
- Whether the player is riding a horse
- World rain intensity
- World time

Screenshot

Custom layers added to provide fire effects, rain effects and time-based-background.

Default profile contains:

- Healthbar (<kbd>Z</kbd> -> <kbd>/</kbd>)
- Experience bar (<kbd>F1</kbd> -> <kbd>F12</kbd>)
- Toolbar (<kbd>1</kbd> -> <kbd>9</kbd>)
- Lit up controls (<kbd>W</kbd>, <kbd>A</kbd>, <kbd>S</kbd>, <kbd>D</kbd>, <kbd>Space</kbd>, <kbd>L-Shift</kbd>, <kbd>L-Ctrl</kbd>) that change colour depending on various things.
- Fire and rain layers.
- Grass block style background that dims during night time.

## Title based profile detection

To support Minecraft, a new type of profile detection has been added that will choose profiles based on the window title. Currently this is only for built-in profiles (user-added profiles have no way of specifying titles).

Title based detection has a lower priority than process name detection. When Aurora is trying to find a profile for the active window, if no profile has been found based on the active process's name, it will check through all the registered titles and test if any of them match the current window.

Titles can be specified in the `LightEventConfig` for an Application under the new `ProcessTitles` property. Like `ProcessNames`, this property is an array of strings however unlike `ProcessNames`, the title strings should be regular expressions. This was done so that versions in the title do not cause a title to be unrecognised. In the case of Minecraft for example, the title is "Minecraft 1.12.2". I chose to avoid testing for a substring as this could cause other programs to be selected when they shouldn't. For example, if it was done so that we test whether "Minecraft" was in the title, any websites with "Minecraft" in the name would also trigger. :+1:


DaemonPhoenix42:
NICE!
Requesting MC Version 1.7.10 Current Forge version for my mod pack (Project Ozone2 v.2.2.2) is 10.13.4.1614

Wibble199:
In future, please put requests for specific Minecraft versions over on the [GitLab issue page for the Minecraft mod](https://gitlab.com/wibble199/aurora-gsi-minecraft/issues) since it's not something the Aurora codebase itself deals with. I shall add MC 1.7.10 to my todo list.

Issue #1259, opened 2018-09-16
Enhancement request: Excluded processes selection improvement


Minotaur1975:
### Expected behavior:

It'd be really great to be able to either "Browse" to an .exe to select it to be an excluded process, or the drop-down list be sorted to alphabetical order. Another option could be to add a profile in the same manner you add a game, but have a the "Allow Aurora to provide lighting effects for" checkbox exclude the process instead if the box isn't checked.

### Actual behavior:

Select arrow next to dropdown box, try to locate the process

### How often does this problem occur?

This problem occurs... (Fill in here) N/A

### Please describe the steps to reproduce this issue:
The issue can be reproduced by doing following steps:

1. Launch process/application to be excluded
2. Alt-Tab back to Aurora
3. Try to locate process from the unsorted list

### What is the version of Aurora you are experiencing this issue with?

My version is... b.466

### Does the previous version of Aurora have this issue?
The previous version, (Fill in the version here), did/did not have this issue.
Unknown

diogotr7:
@Minotaur1975 Thanks for your suggestion. It has been added with PR #1260. If you use GHUB you'll want to grab [this](https://ci.appveyor.com/project/antonpup/aurora/build/0.6.2b-483).

Minotaur1975:
New interface looks great. Thanks!

PR #1260, opened 2018-09-16
Added a dedicated dialog instead of drop down for excluding processes.

Wibble199:
Replaced existing dropdown containing running processes with a sorted, searchable list (with icons!) in a dedicated dialog window. Users can also browse to executable files without needing to run them first.

List showing current running processes.

Option to add a custom (non-running) exe.

Resolves issue #1259.


diogotr7:
Thanks, I never knew i needed this :)

PR #1300, opened 2018-10-16
REST Server feature

boristomas:

- Added button "assign all" to select all available keys (to avoid selecting one by one)
- Added "remove all" because previous
- Added support for listening to REST post with aurora. (there is new "server" tab in settings). Using this feature and ifttt.com (webhooks) users can temporary switch desktop profile for example when new email arrives
- Some other tiny fixes i ran into while figuring out the whole project

**Known issues/To do:**

- We are providing server, that i hope will remain free.
- Please test before final release, currently server is behind IP but I'll put it behind a domain
- Server switching works only with desktop profiles, it would be nice to somehow set active profile even when another app is running profile (for example: i am playing Dota and i got new email, i would like to switch to profile for new email and then take it back to dota) Changing to desktop profile when an app , profile is active does nothing, any idea how to do that?
- It would be awesome if we could hook up aurora on IFTTT but it will cost around 15k usd yearly :)

Peace!
Boris


boristomas:
Hi @antonpup any feedback? tests?

boristomas:
Hi @antonpup any feedback? tests?

Bluscream:
Quick question, can you make shift+click assign only all keyboard keys, not mouse or others

boristomas:
> Quick question, can you make shift+click assign only all keyboard keys, not mouse or others

something like a button: Select only keyboard keys?

Bluscream:
A extra button seems too much to me. Just add something like a `if (Key.Down.Shift) ` to the button onclick handler

PR #1320, opened 2018-11-01
Animation rectangles use origin

Wibble199:
Made the rectangle and filled rectangle rotate about their origins.

This fixes weird behaviour when changing width/height of rotated rectangles. This may also affect any user-made animations but will be minor and easy to fix.


diogotr7:
Thanks!

PR #1338, opened 2018-11-18
Add Animation Layer Triggers

Wibble199:
Added a set of trigger options to the animation layer to allow user's to trigger an animation when a particular value in the game state changes. For example, the keyboard could flash red whenever the player takes damage. Original functionality is still available when the trigger is set to "Always On" (default).

### New options:
- The possible trigger values are:
  - Always on (Same as old animation layer functionality)
  - On value increase
  - On value decrease
  - On value change
  - On boolean become true (Plays once when the value changes from false to true)
  - On boolean become false
  - While boolean true (Plays continuously while the value is true)
  - On key pressed (Plays once when the key is pressed)
  - On key released
  - While key held (Plays continuously while the key is held down)
- For the value triggers, the trigger path is the path to the game state variable to watch for changes.
- For the key triggers, either keybinds can be set or any key can be used ("Trigger for any key").
- For key triggers, the animation can be offset to be over the pressed key.
  - If this is turned on, the animation will be shifted so that the top left corner of the animation canvas is at the center of the pressed button.
  - This allows an interactive layer to be created with a custom animation, opening up for loads more customisation!
  - If, for example, you wanted to create a horizontally expanding rectangle from the pressed key, you would add a filled rectangle track, set "Position X" and "Position Y" to 0, the height to something like 5 and with width at 0. Then make another keyframe and set the width to a higher number (still keeping position X and Y at 0). When a key is pressed and "Translate to pressed key" is checked, the center of the rectangle is translated to the location of the key that was pressed.
- Stack mode dictates what should happen if another trigger is fired while an animation is already playing. Can be set to "Ignore", "Restart" or "Multiple". Multiple replays the animation without cancelling or restarting the current one.
- For key triggers in "pressed" or "held" mode (i.e. not "released" mode), an option is added for whether the animation should be terminated immediately. When false, the animation will keep playing until it finishes.

### New UI:
image

### YouTube demo:
A demo of the animation firing when taking damage (increasing chassis wear) on Euro Truck Sim.
[YouTube demo](https://www.youtube.com/watch?v=GIsC_qCvD0o)

### DIFF
--- a/Project-Aurora/Project-Aurora/EffectsEngine/Animations/AnimationCircle.cs
+++ b/Project-Aurora/Project-Aurora/EffectsEngine/Animations/AnimationCircle.cs
@@ -75,7 +75,7 @@ namespace Aurora.EffectsEngine.Animations
             _duration = duration;
         }
 
-        public override void Draw(Graphics g, float scale = 1.0f)
+        public override void Draw(Graphics g, float scale = 1.0f, PointF offset = default(PointF))
         {
             if (_pen == null || _invalidated)
             {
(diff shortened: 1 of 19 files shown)


Gurjot95:
Great work, can you also add trigger to play animation after key press (could be any key or specific one)?

PR #1341, opened 2018-11-20
Wrapper layer key cloning

Wibble199:
Adds an option to the wrapper layer to allow keys to be cloned. This involves taking the color from a source key and applying it to one or more destination keys. There can be multiple source keys each with their own set of destination keys.

Also quick fix for the "Allow Aurora wrappers to take priority over desktop profile" checkbox so it works as intended.

Thanks to @Gurjot95 for the help with this!


diogotr7:
Thanks :)

PR #1387, opened 2019-01-08
Layer & Profile Copy Improvements

Wibble199:
This commit adds buttons to copy and paste profiles and layers, since many users do not know about the ability to press <kbd>Ctrl</kbd> + <kbd>C</kbd> then <kbd>Ctrl</kbd> +<kbd>V</kbd> commands.

New layer button layout New profile button layout

The profile import feature has also been improved so now JSON files can be imported. This means users will no longer have to manually move the JSON file into the relevant appdata directory and restart Aurora.

In addition, profiles can now be copied from one application to another (e.g. Desktop profiles can be copied to CSGO). Any layers that are not present on the destination application (e.g. layers specific to one app like the CSGO flashbang layer) are removed, but the profile will still copy successfully. Similarlly, this now works when importing JSON profiles - a profile exported from one application can be imported into another application.


Wibble199:
Also, if anyone wants to make better icons than me, please do :)

PR #1432, opened 2019-02-05
Resident Evil 2 (2019) support

srodriguez1850:
I've added RE2 (2019) support via memory reading. Variables included are Health, Maximum Health, Poison, and Rank (dynamic difficulty).

Default profile on preview:
image

**To do:**
- More variables? Plenty of pointers at https://github.com/Squirrelies/RE2REmakeSRT/blob/master/Research/RE2REmake.CT (will probably be maintained for speedrunners)
- Explain rank (dynamic difficulty) on Layer options, in case people don't understand it/confuse it with ending rank (S, A, B, C)?

### DIFF
--- a/Project-Aurora/Project-Aurora/App.xaml.cs
+++ b/Project-Aurora/Project-Aurora/App.xaml.cs
@@ -131,8 +131,11 @@ namespace Aurora
                 Global.isDebug = true;
 #endif
                 Global.Initialize();
-                Directory.SetCurrentDirectory(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
-
+                string path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                Directory.SetCurrentDirectory(path);
+                string logiDll = Path.Combine(path, "LogitechLed.dll");
+                if (File.Exists(logiDll))
+                    File.Delete(logiDll);
                 StringBuilder systeminfo_sb = new StringBuilder(string.Empty);
                 systeminfo_sb.Append("\r\n========================================\r\n");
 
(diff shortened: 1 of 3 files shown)


srodriguez1850:
Will resubmit this PR under the v0.6 branch later today.

srodriguez1850:
https://ci.appveyor.com/project/antonpup/aurora/builds/22129456/artifacts
If you'd like to try.

F0XTHREE:
Tested during game play and everything seems to work as intended. Rank layer displays correctly assigned to the number keys at the top. All 3 effects in the status layer Fine (green) Caution (Yellow) Danger (red) all appear to work with a nice little fade effect for now. 

Suggestions: 
1)A replica animation of the heart monitor moving left to right across the keyboard matching the damage status.
2) Alternatively to suggestion 1, have a slow/medium moving effect going left to right across the entire keyboard (so covering all keys as it moves, could just be a small sector of keys or fill the keyboard as it moves) ending at the mouse. To add to this, the effect could ramp up in speed depending on health status as an experiment.
3) Move the rank layer to the Keypad portion of the keyboard, this may however effect those with "10 key-less" style keyboards so that may not be practical and not everyone would be able to use that layer if moved over.

F0XTHREE:
Seeing some issues after loading into game now. Aurora recognizes the profile is launched however it will stay stuck on the Blue (purple in preview) "Rank Layer" after loading into a save, where as before it would change right away to the health status layer and display the appropriate color for the status. This may be something on my end, ie not getting the update that was pushed to the profile correctly or something along those lines.

srodriguez1850:
Gonna close temporarily since the DLC has been released and changed some pointers around. Will reopen after I've updated them.

Issue #1395, opened 2019-01-13
Steelseries Arctis 5, the right earcup cant be controlled.


ChepaX:
It worked fine, but then I updated firmware of the headset and since than only left earcup is synced to aurora. Any chance of looking into this?

drhead:
This was fixed in Steelseries Engine 3.13.8.

PR #1435, opened 2019-02-08
Optimizations and improvements to the Steelseries support

drhead:

- All requests for Steelseries device lighting changes are consolidated into one event, significantly reducing the CPU usage of Steelseries Engine, especially in comparison to the first implementation of QcK Prism support.
- Adds a 12-zone layout for QcK Prism.
- Adds support for two-zone QcK mousepads like the XL and Cloth, along with a layout to support it. 

**Known issues/To do:**

- This build relies on a newer version of SteelSeries Engine than is publicly available.  Therefore, this should not be merged until the public release of SteelSeries Engine 3.13.7 or later.  This version adds a `show-on-zones:` API that allows mousepad support to not destroy latency and also fixes the Arctis 5 lighting bug (#1395).


drhead:
I'm not entirely sure why there's a conflict with code from the branch I started out with, but lines 182-243 are the ones that need to be removed to resolve the conflict.

PR #1439, opened 2019-02-09
Clone Hero Support

JoeyBallentine:

- Added support for Clone Hero via memory reading
- Added default Clone Hero profile
- Altered MemoryReader to allow 64-bit application support
- Altered MemoryReader to allow selecting a specific module
- Updated other games' memread constructors to account for changes

**Known issues/To do:**

- None


JoeyBallentine:
I'm not sure why these other commits are in this, but this PR is supposed to be entirely for my Clone Hero support.

PR #1440, opened 2019-02-09
Added Clone Hero support via memory reading + Added 64-bit application support to MemoryReader

JoeyBallentine:

- Added support for Clone Hero via memory reading
- Variables are: 
    - Actual Pointer Data:
        - Note streak
        - Star Power active
        - Star Power percent
        - Score
        - Total notes passed 
        - At menu indicator
        - Solo percent
        - Currently pressed frets
    - Inferred Data (after extra calculations)
       - FC indicator
       - Note streak 1x, 2x, 3x, 4x
- Added default Clone Hero profile
- Altered MemoryReader to allow 64-bit application support
- Altered MemoryReader to allow selecting a specific module
- Updated other games' memread constructors to account for changes

**Known issues/To do:**

- Issue: Restarting a song causes the pointers to change. Restarting again fixes it. I attempted to if this by detecting when it was in the restarted state and changing the pointers based on that, but it didn't work. Until we get official GSI this is not fixable.
- Issue: Changing the background image/video breaks pointers, requires game restart. Until we get official GSI this is not fixable.


JoeyBallentine:
Proof that it works via other people:

image

image

image

PR #1451, opened 2019-02-14
Fixed RL Background Layer options being overriden on load

srodriguez1850:
- The Rocket League Background Layer was incorrectly loading the wrong saved property on ShowEnemyExplosion, using the value of ShowGoalExplosion, overriding it on every load.


diogotr7:
Thanks!

Issue #996, opened 2018-02-26
Possible Arduino Support also maybe DMX via Arduino


Wolfieeewolf:
I understand that you have Pi support and are not really interested in adding Arduino support as it's a little difficult due to all the different LED strips. I use a program called Primatik to run my Ambilight setup. It uses the FastLED.io to run the LED's through an Arduino or the like. It's a pretty power by the looks of it. 

It has a way to connect to an API server which if I'm right (I'm not good with coding) should then allow you to use your Aurora API to speak to in which in turn will light up your LED's. 

There is also some, what looks to be easy ways to add Plugins for programs with API's ie iRacing. Which I think someone was asking for support. Looks like you can do some pretty neat things with it. 

I've added a few links if your interested in adding something like this. I know your hard pressed for free time but could be a quick way to get a few more things added to your program. 

https://github.com/psieg/Lightpack

you need https://github.com/dmadison/Adalight-FastLED to get Prismatik to work

http://www.partsnotincluded.com/programming/iracing-plugin-prismatik/ or 
https://github.com/dmadison/Prismatik-iRacing

https://github.com/FastLED/FastLED and http://fastled.io/

I also noted some people asking for DMX support.  Not sure if this could work 

https://github.com/mtongnz/DMX_LED_Strips

https://github.com/OpenLightingProject/ola

I think they are more for controlling the DMX device then sending the info. I'm not overly sure but worth a look if you want. 

I myself use DMX for my Christmas light setup https://www.youtube.com/watch?v=yM74uPYynQE (It's a very old video) and  I now use Vixen 3 http://www.vixenlights.com/ to control it all. I have controllers that cost a lot more then Arudino but can run multible DMX universes and LED strips off one board. You could get quite creative with DMX setups if you have the time and have some really awesome effects as well. Would be great to have Aurora have DMX control but I understand it's not an easy task to get it implemented.

I've added a few more links to some interesting LED controlling programs if your interested as well

https://github.com/kitesurfer1404/WS2812FX

https://github.com/scottlawsonbc/audio-reactive-led-strip

https://github.com/BrianLima/HueHue

I have used most of the links here and they are all pretty easy to setup. Not sure if you can add them to Aurora some how or not. But might help some others who are interested in getting more RGB stuff working along side your program

simon-wh:
[...]

Issue #473, opened 2017-03-02
[Question] Anybody interested in a analog RGB light stripe implementation for Arduino + Aurora?


rkkoszewski:
Hi, yesterday I got myself an Arduino and decided to convert an old decoration LED stripe as a ambi-light for my PC.

Please note that there are two kind of RGB LED stripes. The digital (that support different colors for different areas) and the analog one that has only one global color. 

This implementation is for the second one (analogue). Powered by an Arduino and its serial port connection.

I already have implemented the code running in the Arduino, and the integration with Aurora , and its working very nicely so far. Works even better than my Clevo implementation that requires Admin privileges.  I even implemented color transitions and fading to be computed on the Arduino, for perfect smooth transitions. 

I implemented it on an older version of Aurora, but there should be no issues upgrading it to the latest version. I could do that when I find a free moment.

So I would just like to know if anybody would be interested in this, or if it would be even something worth to integrate into the main Aurora app.

Greetings.

andyman1222:
I will look into this once I get the H-bridges or mosfets connected and working with the Arduino. Do you think you can make something like this but for the Raspberry Pi so it can run alongside digital LEDs? If I had the time I would try doing this myself.

rkkoszewski:
Hi @andyman1222. I don't have a Raspberry Pi nor digital LEDs to implement and test it. If I'm not wrong there is a RPi implementation here that should be working already: https://github.com/antonpup/Aurora/tree/pi-lighting

If @antonpup decides to merge the pi implementation to main then it should be able to work together with the digital RPi and the analog RGB using Arduino.

rkkoszewski:
Here goes the Arduino sketch: https://github.com/rkkoszewski/ArduinoRGBLedDriver
And my implementation in an older version of Aurora:  https://github.com/rkkoszewski/Aurora/tree/dev

Ghanigan:
Would this work with V0.6.0b? I am very interested in this.

antonpup:
@rkkoszewski Umm... the RPi implementation doesn't need to be merged with main. I just wanted to have it on a separate branch so it's easier for people to download and install. My RPi implementation is a script addon, so it will work regardless.

There are two ways to go about your implementation: 1. Merge your Ardino device as part of Aurora and give users options to define the device (ip or whatever you use to connect). 2. You could convert your implementation into a script and PR it to the RPi branch.

But I don't have an arduino not analog lights to test with. So... up to you.

[...]

PR #1454, opened 2019-02-16
Add NodeMCU ESP8266 microcontroller support for driving external LED strips

Wulkop:
This PR introduces a microcontroller compatible implementation of the already implemented LED driver software for the Raspberry Pi.

It allows the user to use a microcontroller (in my case NodeMCUs ESP8266) to receive, deserialize and update the LED status instead of a Raspberry Pi, which is more expensive, bigger and consumes more power.
Since the ESP8266 is WiFi capable this can be done completely wireless.
This new approach is entirely backward compatible with the already existing interface to Aurora which is used to transmit the lighting updates to the Raspberry Pi.

Fixes https://github.com/antonpup/Aurora/issues/473
Fixes https://github.com/antonpup/Aurora/issues/996

* Added a C based and ESP8266 compatible implementation of the already implemented Raspberry Pi software, which:
	* Starts a webserver with REST endpoints on the microcontroller
	* Receives and deserializes the Json lighting updates
	* Sends the new RGB values to the LED strips

- Updated README on how to compile and upload the code to the microcontroller
- Added three new libraries:
	* ArduinoJson (Json library for Arduino microcontrollers)
	* FastLED (LED driver interface which supports up to 14 different LED chipsets (more [here](https://github.com/FastLED/FastLED#supported-led-chipsets))
	* ESP8266 Arduino firmware

### DIFF
--- a/AuroraArduino/AuroraArduino.ino
+++ b/AuroraArduino/AuroraArduino.ino
@@ -1,0 +1,162 @@
+
+#include <ESP8266WiFi.h>
+#include <ESP8266WebServer.h>
+#include <ArduinoJson.h>
+#define FASTLED_ESP8266_RAW_PIN_ORDER
+
+#include "FastLED.h"
+
+#if FASTLED_VERSION < 3001000
+#error "Requires FastLED 3.1 or later"
+#endif
+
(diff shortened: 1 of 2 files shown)


simon-wh:
Looks good, thanks for the contribution!

PR #1458, opened 2019-02-22
Resident Evil 2 (2019) integration

srodriguez1850:
Supersedes #1432.

Memory reading support for Resident Evil 2 (2019 - REmake), along with a default profile and a scanning effect. Pointers maintained at [RE2REmakeSRT](https://github.com/Squirrelies/RE2REmakeSRT/blob/master/RE2REmakeSRT/GameMemory.cs) for speedrunning.

Variables:
* Health
* Max Health
* Poison
* Rank (dynamic difficulty)

Default Profile on Preview:
image


Wibble199:
Regarding your health layer not saving the indicator style, this is because you are not checking if `settingsset` during the SelectionChanged event.

`settingsset` becomes true after the items have been added to the combobox and the selected item of the combobox is programmatically set. By not checking `settingsset` during the SelectionChanged handler, when you call `status_style.Items.Add`, this is causing the selection to change (from nothing to the newly created item), which is then firing the SelectionChanged, which then immediately sets the `_DisplayType` property to be the first inserted value (which is static).

You can see this for yourself if you attach breakpoints to lines 45, 46, 49 and 74: you will be able to see that the handler is executing after 45.

---

Alternatively, you could use the WPF binding feature (which is what DataContext is for) and which is simpler and makes the code more readable since you don't need handlers and extra events cluttering the .xaml.cs file.

srodriguez1850:
Great eye, thanks for the tip!

diogotr7:
Thanks for the contribution!

PR #1469, opened 2019-02-28
Steelseries optimizations and improvements

drhead:

- All requests for Steelseries device lighting changes are consolidated into one event, significantly reducing the CPU usage of Steelseries Engine, especially in comparison to the first implementation of QcK Prism support.
- Adds a 12-zone layout for QcK Prism.
- Adds support for two-zone QcK mousepads like the XL and Cloth, along with a layout to support it. 

**Known issues/To do:**

- This build relies on SteelSeries Engine 3.13.8.  This version adds a `show-on-zones:` API that allows mousepad support to not destroy latency and also fixes the Arctis 5 lighting bug (#1395).


VeraxonHD:
> should not be merged until the public release of SteelSeries Engine 3.13.7 or later.

image

sad

On a serious note though, thanks for this.

drhead:
> > should not be merged until the public release of SteelSeries Engine 3.13.7 or later.
> 
> image
> 
> sad
> 
> On a serious note though, thanks for this.

The update should be released sometime this week, as I am told.  Stay tuned.

simon-wh:
How does this run if someone is using an outdated version of SSE3? Will it just not work as expected?

drhead:
> How does this run if someone is using an outdated version of SSE3? Will it just not work as expected for the extra support with most features that were working before working?

Existing features will work fine, except for 12-zone mousepad support. I did not switch over other devices to the `show-on-zones:` command, so they'll all work as before.

PR #1471, opened 2019-03-01
Game pointers automatic updates on startup

srodriguez1850:
This PR is a first pass at automatically updating all .json files in the "Pointers/" directory on startup. It begins alongside all the startup hooks on App.cs, right before loading applications, and can use either Octokit or an HTTP request to GitHub's raw subdomain to query any branch to get the pointer .json files.

`Utils.PointerUpdateUtils.UpdatePointers("dev")`

The PointerUpdateUtils class was added to the Utils namespace, as to allow future refactors to update the pointers in any part of the UI. The fetch can be relegated to a background task or have the calling thread wait for its completion (not sure if it would ever be used but included for completeness).

As it can use Octokit, unauthenticated requests are rate limited to 60 per day, so unless forced to use Octokit with an optional parameter, it will use the raw subdomain to pull the pointer files. By default, Octokit is only used to pull the list of .json files from the repo directory. Ideally, we'd circumvent the rate limiting by adding an extra file with the list of supported games, so that on an HTTP request to this file, we get all the pointer files we need. Will keep this drafted until discussion.


Project-Aurora/Project-Aurora/Utils/PointerUpdateUtils.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            if (!backgroundTask) Task.Run(() => FetchLocalPointers(branch)).Wait();
+            else Task.Run(() => FetchLocalPointers(branch));
+        }
+
+        private static async Task FetchAllPointers(string branch, bool useOctokit)
+        {
+            // Update pointer files in Aurora/Pointers/
+            string pointerPath = Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName);
+            pointerPath += @"\Pointers\";
```
Wibble199: Minor suggestion: Use `string pointerPath = Path.Combine(Global.ExecutingDirectory, "Pointers");` This is what the pointer readers use to resolve their path.

srodriguez1850: Thanks!

srodriguez1850:
Startup method is now UpdateLocalPointers from the "dev" branch (`Utils.PointerUpdateUtils.UpdateLocalPointers("dev")`). This is so if a new game is merged into dev, it won't automatically be pushed to everyone's client, but rather wait until the next update to start getting updates. There's probably a good way to get the memread-supported games within Aurora that I'm missing.

srodriguez1850:
New functionality of `Aurora.Utils.PointerUpdateUtils` listed as follows:
- **Marking games for pointer update on application load**
    - To prevent users from getting pointer files for games that are not present in their build but present in the repo, `MarkAppForUpdate(string pointerFile)` can be called in the constructor of their respective Application.cs file. The parameter is the name of the pointer file to be pulled. That way we can update the repo and merge new games, and the user will only pull pointers they need.
- **Exposed two methods for updating pointers**
   - `FetchPointers(string branch, bool useOctokit)` pulls only the pointer files marked for update, as mentioned above.
   - `FetchDevPointers(string branch, bool useOctokit)` pulls everything from the repo, even for games not present in the user's build. Use sparingly.
    - Leave `useOctokit` as false, as every file makes an API request and can hit 60 pretty fast as the games list grow. Included it just if reaaaaally needed.
    - The task wrappers were removed, so they'll need the `Task.Run(Action)` method to complete. Devs can program as they'd like.

This also adds the `MarkAppForUpdate(string)` method to all memread game integrations. Any future integrations must add this method if they want pointer/file updates.

simon-wh:
I'd suggest having a look at the `Global.configuration.GetDevReleases` and consider if that may be helpful or not for determing the branch to get updates from? In theory any updated pointers on the dev branch should be fully sorted, but maybe you may find that using this option may help. Otherwise, looks great!

simon-wh:
I think it may be a bad idea to always update from dev, we don't want a case where we decide to do breaking changes on a pointer file and it suddenly gets pushed to everyone. I think it may be better to get updates from master and all pointer updates that have been verified can be cherrypicked into the master branch for the clients to update from.

Project-Aurora/Project-Aurora/Settings/Configuration.cs
```diff
@@ -395,6 +395,7 @@ public class Configuration : Settings
         public BitmapAccuracy BitmapAccuracy { get { return bitmapAccuracy; } set { bitmapAccuracy = value; InvokePropertyChanged(); } }
 
         public bool updates_check_on_start_up;
+        public bool pointer_updates_check_on_start_up;
```
simon-wh: This should be setup as a property similar to BitmapAccuracy, not a variable like the rest of the old parts of Configuration

Project-Aurora/Project-Aurora/Settings/Control_Settings.xaml
```diff
@@ -277,8 +277,9 @@
                 <Grid>
                     <Button x:Name="updates_check" Content="Check for Updates" HorizontalAlignment="Left" Margin="10,31,0,0" VerticalAlignment="Top" Click="updates_check_Click"/>
                     <CheckBox x:Name="updates_autocheck_on_start" Content="Automatically check for updates on program start" HorizontalAlignment="Left" Margin="10,56,0,0" VerticalAlignment="Top" Checked="updates_autocheck_on_start_Checked" Unchecked="updates_autocheck_on_start_Checked"/>
+                    <CheckBox x:Name="pointer_updates_autocheck_on_start" Content="Automatically check for pointer updates on program start" HorizontalAlignment="Left" Margin="10,76,0,0" VerticalAlignment="Top" Checked="pointer_updates_autocheck_on_start_Checked" Unchecked="pointer_updates_autocheck_on_start_Checked"/>
```
simon-wh: This should be binding to the Property in Configuration, not using event handlers, see L169

srodriguez1850:
Is the plan changing all the variables in Settings to WPF bindings? It seems that if an option is changed through binding, it does not save unless another UI element (that is not binded) calls `ConfigManager.Save()`. I think it's because none of the WPF property callbacks actually calls `Save()` and writes out to *Config.json*.

https://github.com/antonpup/Aurora/blob/1f9eace5278ec0827700e7f7ab65e956f60d93e1/Project-Aurora/Project-Aurora/Settings/ApplicationProfile.cs#L14-L19

simon-wh:
Ideally all variables in Configuration will be switched over to properties which work with INotifyPropertyChanged. With that, to make it save whenever anything is changed, all is required is to subscribe to the PropertyChanged event and call save. 

I've changed them up slightly in device-layout-rewrite, but I'll update them all to this once this is merged in.
https://github.com/antonpup/Aurora/blob/feature/device-layout-rewrite/Project-Aurora/Project-Aurora/Settings/Configuration.cs#L366

simon-wh:
Thanks very much for the contribution!

PR #1483, opened 2019-03-04
HTTP State Integration Plugins

Wibble199:
This pull request adds in the ability to use C# or DLL files as state plugins which add new game states that can be accessed from any application. (I did consider making it application-specific but there are some use cases you want it for all, plus it was proving to be more complex than making it global).

The states can be updated by making a HTTP request to the Aurora listener with a provider node that has a name property matching the ID of the plugin. Note that the ID is striped of any slashes and spaces are turned into underscores.

All plugin game states appear under a plugins node, followed by their ID. For example, a plugin with an ID of "MyGSI" and a field "MyField" would have the path `Plugins/MyGSI/MyField`.

An example of how this could be useful is allowing users to add support for their own applications without needing to create a separate branch to make changes and potentially then have an outdated version of Aurora. The reason I developed this was so that I could make integration with the LSPDFR mod for GTA V, but I couldn't really add this as a separate game (since it is still gtav.exe) and many GTA V players will not play this mod so it isn't fair to change the GTA V profile for my own needs. This way, I can write my plugin once and then not have to worry about keeping a private custom branch up to date or whatever.

Documentation on how exactly to create plugins will be added to the GitBook guide soon.

---

Secondly, a new contol has been added to pick the game state variables. It separates the list into sections, only showing one parameter at a time, for example "CurrentMinute" and "CurrentSecond" are both underneath the "LocalPCInfo" node.

The new GSI selection

This is particularlly useful for games with a lot of GSI variables or when the user has a lot of plugins providing new parameters. (And before you ask, yes I did try using a tree but it wasn't easy to convert the parameters to the right tree structure or to actually programmatically set the tree's selected item, nor is it as pretty).

---

An additional noteworthy change is to add an `AutoNode` class that works the same as the `Node` for game states but attempts to automatically populate the fields/properties when it recieves JSON. This eliminates the need for manually doing this cases where the fields in the node simply mirror the ones from the JSON. A good example of this is in use is the Minecraft PlayerNode.

The profile/application script loading has now also been rewritten slightly to use the new utility code for external scripts/assemblies. This change does mean that compiled DLLs can be used for profile scripts if people want to do it that way for whatever reason. Also, python scripts _should_ work with the GSI plugin system (untested, but it works properly with the scripts layer). This utility code makes it _incredibly_ easy should we wish to add other plugins in the future, such as doing profiles as plugins, overridables as plugins, etc.


Wibble199:
I've been thinking about this more recently, and I'm wondering if it would be better to build on/rewrite the existing plugin system to allow generic plugins that can do many things (such as HTTP states, registering new layers, registering devices, etc.). Thoughts?

simon-wh:
> I've been thinking about this more recently, and I'm wondering if it would be better to build on/rewrite the existing plugin system to allow generic plugins that can do many things (such as new application profiles, global HTTP states, registering new layers, registering devices, adding new evaluatables etc.). Thoughts?

That's what my intention was with the Plugin system, although I didn't actually finish it. Ideally I'd like it to be such that the core of Aurora is just the system and all applications, devices, layers, etc are added through plugins. You can see from the way LightingStateManager is designed that it was something I had in mind while designing the class, I tried to decouple the Applications and Layers as much as possible from the rest of the code so that to add one is just a matter of adding one line or using a plugin.

With that said, my intention for the kind of thing that you've done here was to have the plugins have the ability to add game states that can be accessed globally and ones specific to applications added in the plugin.

PR #1490, opened 2019-03-09
Minor Overrides System Improvements

Wibble199:
This PR attempts to fix a few issues that have been found with the Overrides System. Current known and fixed bugs are:

* [x] Game State dropdowns that are children of and/or expressions being empty
* [x] ~Changing the overrride logic does not trigger a save and so sometimes you can lose progress if you close Aurora immediately after editing the logic.~ According to [this Discord message](https://discordapp.com/channels/427609830073696258/428606931700023299/554318942823251988), it's not an issue with the overrides system in particular, so skipping it in this override fix PR.
* [x] Add exclude keys/regions to all Layers
* [x] Add all properties to the LayerEffectConfig editor
* [x] Improve audio visualizer support (Add secondary color, etc.)
* [x] Add current/maximum value to percent Layers
* [x] Added a **temporary** `LogicOverrideIgnoreProperty` attribute for layer handlers so specify that one or more of their properties should not be shown in the overridable property list. This should be **temporary** until we are able to resolve the layer handler properties' inheritance, which I recommend is done next time there is a change planned that will likely break existing profiles anyways (such as the device layout rework perhaps?).

### DIFF
--- a/Project-Aurora/Project-Aurora/Controls/Control_FieldPresenter.xaml.cs
+++ b/Project-Aurora/Project-Aurora/Controls/Control_FieldPresenter.xaml.cs
@@ -85,15 +85,8 @@ namespace Aurora.Controls {
             { typeof(RealColor), bind => new ColorPicker{ ColorMode = ColorMode.ColorCanvas }.SetBindingChain(ColorPicker.SelectedColorProperty, bind, new RealColorConverter()) },
 
             // Gradient colour
-            { typeof(Settings.LayerEffectConfig), bind => {
-                // Still needs some work to become properly useful since the animation effect and speed and stuff are stored
-                // in the layer effect config in addition to the brush, so this causes those values to also get overridden.
-                var canvas = new ColorBox.ColorBox();
-                var src = (Control_FieldPresenter)bind.Source;
-                canvas.Brush = ((Settings.LayerEffectConfig)src.Value).brush.GetMediaBrush();
-                canvas.BrushChanged += (sender, e) => { ((Settings.LayerEffectConfig)src.Value).brush = new EffectsEngine.EffectBrush(e.Brush); };
-                return canvas;
-            } },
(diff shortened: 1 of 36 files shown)


gitmacer:
### here are some more think that should be changed:

Remove Primary/Secondary Color Overwriting for following Layers:
-Gradient Layer
-Gradient Fill Layer
-Image Layer
-Percent (Gradient) Effect Layer
-Glitch Effect Layer
-Animation Layer

Also:
-Script Layer: Remove Primary Color and Affected Keys
-Remove Affected Keys from Fill Layers
-Add Exclude Keys/Regions to all Layers
-Add Gradient overwriting to Percent (Gradient) Effect Layer 
-Gradient logic (LayerEffectConfig) should not overwride other variables(Effect Speed... should also be overwritable)
-Improve Audio Visualizer support (no secondary Color...)

Effect Layers:
-Add Current/Maximum Value

Wibble199:
The reason that many layers have additional properties is because the layer handler properties inherit from the `LayerHandlerProperties2Color` class which is exposing the primary/secondary properties. While would be possible with some layers to remove this relationship and inherit from the `LayerHandlerProperties` class instead, there are many layers that use this directly (rather than creating their own properties and extending) which means that changing the names or types of the property classes will **break existing profiles**. If there is ever a release planned that will knowingly break profiles for other reasons then I will consider tidying up this, however I think it currently does not warrant breaking profiles for the sake of a few properties being present when they shouldn't be.

gitmacer:
ok, but If it is not possible to easily remove these properties it should still be possible to fix following things:

-Add Exclude Keys/Regions to all Layers
-Add Gradient overwriting to Percent (Gradient) Effect Layer
-Gradient logic (LayerEffectConfig) should not overwride other variables(Effect Speed... should also be overwritable)
-Improve Audio Visualizer support (no secondary Color,...)

Effect Layers:
-Add Current/Maximum Value

gitmacer:
Freestyle selection is affecting others when at the same position. https://cdn.discordapp.com/attachments/428606931700023299/564876515037151235/freestyle_bug.gif

gitmacer:
Ok this bug is not related to this pr, because the Skype overlay has the same bug.

PR #1492, opened 2019-03-10
Feature/debug windows always on top

gitmacer:
This feature is especially useful if you have only one Monitor or you want to show a bug and the Debug window need to be in front of an active Window. :)

- Add options to enable Top Most (Always on top) for Debug Windows

Demo:
grafik

**Known issues:**

- ~~You need to reopen the Debug window that the option takes effect~~
- Full-screen Games are still in front

- Save issue not really fixed you can still lose Data that was changed within 300 ms


Project-Aurora/Project-Aurora/Settings/Configuration.cs
```diff
@@ -440,6 +440,10 @@ public class Configuration : Settings
         public VolumeOverlaySettings volume_overlay_settings;
         public SkypeOverlaySettings skype_overlay_settings;
 
+        //Debug Settings
+        public bool BitmapTopMost;
+        public bool GSIHttpDebugTopMost;
```
simon-wh: These should be set up as properties that work with INotifyPropertyChanged (see the HighPriority property)

Project-Aurora/Project-Aurora/Settings/Control_Settings.xaml
```diff
@@ -304,8 +304,10 @@
                     <TextBlock x:Name="textBlock" HorizontalAlignment="Left" Margin="10,10,0,0" TextWrapping="Wrap" Text="Current Keyboard bitmap:" VerticalAlignment="Top"/>
                     <Image x:Name="debug_bitmap_preview" HorizontalAlignment="Left" Margin="152,10,0,0" VerticalAlignment="Top" MaxWidth="400" MaxHeight="400"/>
                     <Button x:Name="btnShowBitmapWindow" Content="Show Bitmap Window" HorizontalAlignment="Left" Margin="10,31,0,0" VerticalAlignment="Top" Click="btnShowBitmapWindow_Click"/>
-                    <Button x:Name="btnShowLogsFolder" Content="Show Logs Folder" HorizontalAlignment="Left" Margin="10,56,0,0" VerticalAlignment="Top" Click="btnShowLogsFolder_Click"/>
-                    <Button x:Name="btnShowGSILog" Content="Show GSI HTTP requests" HorizontalAlignment="Left" Margin="10,81,0,0" VerticalAlignment="Top" Click="btnShowGSILog_Click" />
+                    <CheckBox x:Name="BitmapTopMost" Content="Always on top" HorizontalAlignment="Left" Margin="10,56,0,0" VerticalAlignment="Top" Unchecked="bitmap_TopMost_Checked" Checked="bitmap_TopMost_Checked"/>
```
simon-wh: These Checkbox's should be binding to the Property in the Configuration, not using event handlers, see Line 169

Project-Aurora/Project-Aurora/Settings/Configuration.cs
gitmacer: it now uses propertys.

Project-Aurora/Project-Aurora/Settings/Control_Settings.xaml
gitmacer: it is now using binding to properties.

Project-Aurora/Project-Aurora/Settings/Configuration.cs
```diff
@@ -441,8 +441,14 @@ public class Configuration : Settings
         public SkypeOverlaySettings skype_overlay_settings;
 
         //Debug Settings
-        public bool BitmapTopMost;
-        public bool GSIHttpDebugTopMost;
+        private bool bitmapTopMost;
+        [JsonProperty(PropertyName = "bitmap_debug_topmost")]
```
simon-wh: The JsonProperty renaming should only be used for compatibility with the old name of a property in Configuration. It shouldn't be used for new properties, the names used in the json should be the same as the Property unless there's a good reason to change it to something else

gitmacer: no longer doing JsonProperty renaming

PR #1514, opened 2019-03-22
Feature/better audio layer

gitmacer:

Fix issue reported over Discord:
[Link](https://discordapp.com/channels/427609830073696258/428606969125797888/554049334497247260)
[Link](https://discordapp.com/channels/427609830073696258/428606969125797888/554053799015350365)

- Set CompositingMode for the Background Option to SourceCopy. Needed if you want to use transparent as color and enable the background.
- Add Background always enabled option
- (try to) make the Layer look better
- Add default _ScaleWithSystemVolume value

Demo:
Old:
old
new:
new
demo_gif

**Known issues/To do:**

- Old Dim background on Sound option will not get converted to new option

**More Feature Ideas:**
- Freestyle option for the Equalizer would be great and needed after the device layout rewrite
- Frequencies presets?


diogotr7:
People on the discord have suggested that the default frequencies from this layer should be changed to the ones on my profile because they look better. I think this PR would be a good opportunity to include them. Opinions?

simon-wh:
> People on the discord have suggested that the default frequencies from this layer should be changed to the ones on my profile because they look better. I think this PR would be a good opportunity to include them. Opinions?

That sounds like a good idea. I feel like the frequency option should be improved in general anyway, to make it more usable for the end user. Some audio visualisers I've looked at don't ask you for particular frequencies, but instead ask for min/max and the number of samples/bars. With how it is now, most people won't understand how to actually use the option. Maybe we could have Simple & Advanced modes?

Project-Aurora/Project-Aurora/Settings/Layers/EqualizerLayerHandler.cs
```diff
@@ -120,9 +132,10 @@ public override void Default()
             _EQType = EqualizerType.PowerBars;
             _ViewType = EqualizerPresentationType.SolidColor;
             _MaxAmplitude = 20.0f;
-            _DimBackgroundOnSound = false;
+            _ScaleWithSystemVolume = false;
+            _BackgroundMode = EqualizerBackgroundMode.Disabled;
             _DimColor = Color.FromArgb(169, 0, 0, 0);
-            _Frequencies = new SortedSet<float>() { 60, 170, 310, 600, 1000, 2000, 3000, 4000, 5000 };
+            _Frequencies = new SortedSet<float>() { 50, 95, 130, 180, 250, 350, 500, 620, 700, 850, 1200, 1600, 2200, 3000, 4100, 5600, 7700, 10000, 10001, 10002 };
```
diogotr7: The last few values (10k+) are my personal preference to move the other bars a bit to the left. I think the last 2 frequencies should be removed from the default ones.

PR #1527, opened 2019-04-04
Made Profile Keybind Box wider

### LINKED ISSUE #1525: Profile Keybind not Entirely Visible
### Expected behavior:
Entire keybind should be visible in the profile switching keybind box. This could be fixed either by making the box wider or by making it scrollable.

### Actual behavior:
Only part of the keybind is visible in the box when the keybind is long enough.

Whenever a long keybind is added to a profile.

Add a long keybind to a profile (Ex. Ctrl+Shift+Alt+Y)

beta v0.6.4b-633

I assume that the previous versions with support for profile switching also had this issue.

diogotr7:
Fixes #1525


diogotr7:
I used Ctrl + Alt + Shift + Key and it fit, I don't know if there are any longer key sequences but this should ne fine.

Issue #1419, opened 2019-01-29
Support for Corsair Fans (Lightning Node Pro)


Kalbi1:
I have a problem with my RGB Setup.
First of all, I really love your software! It's amazing that I can use all my devices with only one software.

**My Setup:**
Logitech G810
Corsair Polaris
Corsair LL Fans
Logitech MX Master 2s

My issue is that my Corsair LL Fans turn off, if I launch Aurora, because of the Corsair Polaris support. Is there any way to keep my fans working because they turn off... or do you want to support RGB Fans, LED Strips and other things? I think it would be really cool. The Corsair Polaris support is amazing by the way. Can you do a similar implementation with PC Fans?

I think it is a common problem for many users of Aurora RGB and many people would love to control their fans and RGB Stripes with Aurora too, because if you have other Corsair Products you cant use Aurora because the Fans turn off :(.

Best Regards from Germany
Henrik

Gurjot95:
Are you fine with mouse color supporting all other peripherals color such as corsair fans, graphic cards etc? because I can try to implement that temporary till rewrite is done. The rewrite is coming soon which will support more peripherals with their individual addressable LED. Right now, I think broadcasting mouse color to all other peripherals would do the job.

JMZ10:
I don't have the fans but I have the corsairs strips. I wouldn't mind that as a temporary fix, as right now I have to disable corsair support on aurora because the strips just black out which is annoying. I don't if its possible alongside the temp fix, but maybe for those with more corsair gear (i only have strips) might appreciate an option to let iCue control the addressable stuff while aurora still controls everything else. Again I'm not sure if that's possible.

Kalbi1:
I think as a temporary solution it would be alright if you could just adjust that the other devices that are not supported simply shine in the color in which the keyboard or mouse is also lit. So no individual control for each LED but just one color for all LEDs.

Then you could at least use Aurora with a mouse pad or a mouse without the fans turning off, because that is really sad if you have spent 90 Euros for 3 Fans and they turn off xD.

Gurjot95:
I have added the PR which adds support for most peripherals and follows mouse color. There is now also option in Device Manager to use SDK in Shared Mode with iCue so the unlit Corsair devices from Aurora can be controlled from iCue
https://ci.appveyor.com/project/antonpup/aurora/builds/22008450/artifacts

Kalbi1:
It works very well! Thank you ;).

[...]

PR #1530, opened 2019-04-06
 Corsair Peripherals Support (including RAM, Fans, LED Strips), added Shared Mode

Gurjot95:
#1419 

-  Adds support for most corsair peripherals such as RAM, Lighing Node, Commander Pro etc which follows mouse color
- Added device option to whether to run SDK in shared mode with iCue and option to disable each peripheral
- Optimized whole corsair device code (needs testing so all corsair devices are working fine)
- Updated SDK, this time for real
- Should fix RAM issue from last PR (thanks @srodriguez1850)
- Fixes Harpoon Mouse issue

**Known issues/To do:**

- Needs extensive testing


zejjnt:
How can I test this? Since I can find no executable, how do I compile it?

diogotr7:
@zejjnt you have a premade build here https://ci.appveyor.com/project/antonpup/aurora/builds/23648968/artifacts

zejjnt:
Well this is kind of a mixed bag, but there is clearly progress. The mouse pad and keyboard are in excellent sync and that's just awesome.
But from what I've seen, the fan colors are set based only on one single spot; the mouse logo, correct? Because while they do blink every now and then it is always the same color and there is no fading whatsoever, either all fans on or off in the same static color.
I'm guessing you're going to expand on this feature in the future but I still have to point out that color gradients and different fans for different audio frequencies is the ultimate goal, isn't it? :)

Unfortunately my RAM is still doing its default pattern from iCue despite iCue saying that they are controlled via the SDK. My SteelSeries Arctis 5 is not in this equation at all, they are just doing the default rainbow fading.

Great work anyhow :)

Axodious:
For me the fans are responding to my typing fine (in desktop mode).

I did notice a possible major problem, when Aurora is opened the fans disappear from iCue which seems to remove the fan speed control (I think). Woke up this morning with my computer pretty warm, went to iCue and noticed the fans not in the Commander Pro, uninstalled Aurora, rebooted and the fans were back in iCue.

zejjnt:
> For me the fans are responding to my typing fine (in desktop mode).
> 
> I did notice a possible major problem, when Aurora is opened the fans disappear from iCue which seems to remove the fan speed control (I think). Woke up this morning with my computer pretty warm, went to iCue and noticed the fans not in the Commander Pro, uninstalled Aurora, rebooted and the fans were back in iCue.
> 
> _EDIT_
> It looks like it might not be just Aurora causing the fans to stop showing RPM in iCue, a while after the reboot the fans disappeared by themselves. I did notice that if they are there and I start Aurora they disappear right away.
> 
> _EDIT 2_
> I found this post regarding the Commander Pro: http://forum.corsair.com/v3/showthread.php?t=181855#9
> 
> Apparently you can only have 1 connection to it at a time, so when Aurora is taking over it's stopping iCue from monitoring and adjusting any fans plugged in as well. Not sure there's a workaround for Aurora other than adding it's own form of fan speed controls.
> 
> _EDIT 3_
> One workaround is to plug the fan controls directly on the motherboard and only plug the RGB cable into the Commander Pro. Don't have to worry about iCue controlling the fan speed at all this way.

That might explain why iCue often starts on another profile than what it was set to when I shut down my computer maybe.
And about the only one device-thing I noticed something in the Device Manager related to what you are writing;
image
As I have said I have two Lightning Nodes, one Commander Pro, 2 LL-fans, 3 ML-fans and one MM800RGB mouse pad but only the mouse pad is showing.
Meanwhile I have a Logitech mouse (G502) that is not detected, SteelSeries Arctis 5 which are detected but only change color sometimes - and a Razer Chroma Blackwidow v2 which somehow registers as a lot of stuff that it isn't.

I don't know if this is any help to you but I thought why not give the info I have...

maestroraymond:
I will be testing this out with the Lighting Node Pro over the weekend.
Have been flawlessly using Aurora with a Steel Series Rival (flickering aside, workaround: switch preferred device to another mouse), Corsair K70 + some additional devices every now and then.
I'm new to GitHubbing so don't know if it's the right place for these kinds of notes, but hey ho, just my input on the great project that is Aurora:
- Devices could be separated and moved/dragged/swapped to the desired position in the manager (eg. left/right handed mouse users)
- As far as the LightingNode and similar LEDs go, I saw another pull request about Hue+ devices, would it be possible to implement a similar approach? (following the keyboard lighting, or assignable lighting like LedStrip1 copy lighting of f1-f12, LedStrip 2 esc-ctrl)

diogotr7:
@maestroraymond 
The approach of making the light strips follow the keyboard like the nzxt hue PR is most likely possible.

Your idea of dragging devices around the screen is partially working on another branch, so that is the way Aurora is going to work in the future. Also proper support for case lighting and keypads for example is planned along with that change.

Your githubbing is just fine :)

maestroraymond:
Thanks for the kind words @diogotr7.
I'm coming back to coding and I'll start playing with some other repo's as currently the only way to make a node pro work somewhat correctly as 'ambilight' feature (apart from following the mouse or any other pre-set peripherial is syncing/slaving the individual leds to specific keys on the keyboard. I am currently testing this with DarthAffe's RGB.NET/RgbSync+ alongside Aurora.
Will update if that works correctly!

Gurjot95:
@maestroraymond @zejjnt @Axodious 
Guys please check the new PR build #1580 which adds Corsair support using RGB.NET. Let me know in that thread if it works!

Axodious:
I'm away for a week I'll check when I get back.

zejjnt:
NOW we're talking! Everything lights up now, SteelSeries, Corsair and Razer correctly and in sync. I think the SteelSeries are still controlled by Engine but it works anyway (just without automatically matching colors).
The fans only light up in one color but that is more than enough to impress, it's quite fluent when you set "Fine" instead of "Best".
Great work people! <3

Gurjot95:
Great, please any more updates in #1580 thread as I am closing this for now. Thanks!

PR #1534, opened 2019-04-10
fix of not re enabling "Use Freestyle instead" checkbox

gitmacer:
It seems like it was originally meant to disable the controls when the use Freestyle instead checkbox is checked.

- fix of not re enabling "Use Freestyle instead" checkbox
- 
- 

**Known issues/To do:**

- wrong status at startup
- 
- 
Exclude Keys bug


gitmacer:
What? AppVeyor? Why do you not want to build my pr?

gitmacer:
Closed because AppVeyor does not want to build it.

PR #1539, opened 2019-04-14
Added support for NZXT Hue+ and KrakenX

### LINKED ISSUE #132: Can someone link NZXT Hue+ with all my device
i really need it so i can match my pc case light .. it would be very helpful to me.. sorry, im new here..

diogotr7:
Fixes #132 

- Adds support for the Hue+ lighting controller and the KrakenX line of CPU coolers

**Known issues/To do:**

- Currently follows keyboard keys for the lighting. It would be ideal to have it as its own independent device in the device rewrite. I'm not sure what the best way would be to add support for it now, as adding 80 device options and/or keys and a layout doesn't seem practical.
- Causes flickering, cause is currently unknown. Might be a result of sending too many updates. Should probably not merge until this is fixed as it may cause hardware problems.
- **Causes aurora to use much more memory and cpu than usual.**

Also thanks @akmadian for his help :)


akmadian:
Devices could be broken down from physical devices like a whole Hue+ or KrakenX to individual channels, or even (would take a fair amount more coding work) individual strips/ fans, and even (if you want to go that far) individual LEDs. The support for this is already in NZXTSharp, it would just be up to you to figure out exactly *how* that implementation might work, and if you're up for doing all the work that would come with it.

akmadian:
Also, since you have support for channel 1, I'm not sure why you couldn't add support for both? Should be just a matter of applying the effect to `HuePlus.Both` instead of `HuePlus.Channel1`

Project-Aurora/Project-Aurora/Devices/NZXT/NZXTDevice.cs
```diff
@@ -0,0 +1,198 @@
...
+            if (e.Cancel) return false;
+
+            try
+            {
+                foreach (KeyValuePair<DeviceKeys, Color> key in keycolors)
+                {
+                    if (e.Cancel) return false;
+                    if (key.Key == DeviceKeys.Peripheral_Logo)
+                    {
+                        DeviceLoader.KrakenX.ApplyEffect(DeviceLoader.KrakenX.Logo, new NZXTSharp.Fixed(new NZXTSharp.Color(key.Value.R, key.Value.G, key.Value.B)));
```
akmadian: Add ? null checks to devices since the DeviceLoader class doesn't do this :/

Project-Aurora/Project-Aurora/Devices/NZXT/NZXTDevice.cs
```diff
@@ -0,0 +1,198 @@
...
+                             key.Key >= DeviceKeys.A && key.Key <= DeviceKeys.SEMICOLON ||
+                             key.Key >= DeviceKeys.Z && key.Key <= DeviceKeys.PERIOD)//40 keys
+                    {
+                        huepluscolors.Add(key.Value.G);
+                        huepluscolors.Add(key.Value.R);
+                        huepluscolors.Add(key.Value.B);
+                    }
+                }
+
+                DeviceLoader.HuePlus.ApplyEffect(DeviceLoader.HuePlus.Both, new NZXTSharp.Fixed(huepluscolors.ToArray()));
```
akmadian: Add ? null check operators to ApplyEffect methods

Project-Aurora/Project-Aurora/Devices/NZXT/NZXTDevice.cs
akmadian: Add ? null checks to devices since the DeviceLoader class doesn't do this :/

Project-Aurora/Project-Aurora/Devices/NZXT/NZXTDevice.cs
```diff
@@ -153,8 +157,21 @@ public bool UpdateDevice(Dictionary<DeviceKeys, Color> keycolors, DoWorkEventArg
                     }
                 }
 
-                DeviceLoader.HuePlus?.ApplyEffect(DeviceLoader.HuePlus.Both, new NZXTSharp.Fixed(huepluscolors.ToArray()));
-                DeviceLoader.KrakenX?.ApplyEffect(DeviceLoader.KrakenX.Ring, new NZXTSharp.Fixed(krakenringcolors.ToArray()));
+                if (!huepluscolors.Equals(hue_device) || forced)
```
Wibble199: This doesn't check if the two lists are equal, it checks if they are referencing the same object. Since you are creating `huepluscolors` at the start of this method, the references will never be the same. To check if the lists are equal a different method such as `Enumerable.SequenceEqual` should be used.

I've created a [.NET Fiddle](https://dotnetfiddle.net/a6ohls) illustrating this in this context.

akmadian:
Is there any reason that you have the `UpdateValueAt` method instead of just assigning to that index like `arr[index] = value`?

diogotr7:
Wibble suggested that method. It makes it so any changes to the array cause the update to the hardware to be sent. if the array of colors stays the same, it doesnt update the device.

akmadian:
Ahhh, yeah that's actually pretty clever ;)

PR #1541, opened 2019-04-15
Updated League of Legends patcher with new game path

### LINKED ISSUE #1540: League of Legends LightFX no longer work after new patch
A few days ago I was notified by the game launcher, that the developer (Riot Games) was moving some game files around. Since then, the LoL wrapper lighting has not worked. I've been using it for about a month and a half now with no problems, I've had it working in this current game patch, but after that notification, it hasn't worked once. Automatic patching in Aurora no longer seems to work. When I click "Patch League of Legends" nothing happens (this is also new since the launcher notification). I manually patched both releases in my game folder, but to no effect. I checked both directories manually and the LightFX.dll file is located in the "deploy" folder in both. Other than patching manually instead of automatically, I changed nothing else. Later I also tried creating a new LoL profile and kept it default and it didn't work either.
I have a Wooting Two keyboard.

### Expected behavior:
When in-game in League of Legends, the LoL Aurora profile should change the keyboard lighting in the wrapper layer depending on the in-game HP.

### Actual behavior:
The wrapper layer stays dark (off) as if I wasn't in-game or the game wasn't properly patched, etc. It doesn't turn on.

This problem occurs every time. 

1. Patching League of Legends (with LightFX enabled in the config)
2. Starting the game

My version is... beta 0.6.4b-624-lkowurnn
This problem did not exist in this same version until recently.

I don't know, I haven't used any other version.

### Please include any logs, screenshots, or files that are related to this issue:
Logs:
[2019-04-14 19.49.56.log](https://pastebin.com/4RJGsrSm)
[2019-04-11 19.55.18.log](https://pastebin.com/fEbWH5ED)

Screenshot:
[game directories for patching](https://i.imgur.com/r1fbvvB.png)

diogotr7:
Fixes #1540

Riot Changed the path of the game executable so aurora doesnt need to apply the patch every game update. This PR applies that change to the path

### DIFF
--- a/Project-Aurora/Project-Aurora/Profiles/LeagueOfLegends/Control_LoL.xaml.cs
+++ b/Project-Aurora/Project-Aurora/Profiles/LeagueOfLegends/Control_LoL.xaml.cs
@@ -68,35 +68,20 @@ namespace Aurora.Profiles.LeagueOfLegends
 
                 if (!String.IsNullOrWhiteSpace(lolpath))
                 {
-                    lolpath = Path.Combine(lolpath, "RADS", "solutions", "lol_game_client_sln", "releases");
-                    DirectoryInfo[] dirs = new DirectoryInfo(lolpath).GetDirectories();
-                    if(dirs.Length != 0)
+                    lolpath = Path.Combine(lolpath, "Game");
+                    if (Directory.Exists(lolpath))
                     {
-                        string latestversion = dirs[dirs.Length - 1].ToString();
-                        lolpath = Path.Combine(lolpath, latestversion, "deploy");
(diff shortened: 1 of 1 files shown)


diogotr7:
I'll use this PR also to ask a question: should the "patch manually" buttons assume the user chooses the root directory of the game, the game where the dll should be placed, or account for both and adjust accordingly? I'm asking because for Factorio, the manual patch button assumes the user chooses the root directory, appending "bin" and "x64" to it. In other profiles this isn't the case. Should this be changed?

diogotr7:
As per #1540 im closing this as i dont think its fixable

PR #1544, opened 2019-04-18
support for new cooler master SDK25 (CK530,MM830,MM850)

OneB1t:
new devices for cooler master


simon-wh:
Thanks for the contribution!

PR #1560, opened 2019-05-05
Fixed logitech wrapper registry path not being applied corectly

diogotr7:
This PR makes the Patch Logitech button write the correct dll path as the **default value** for the registry key. 
The way this was done previously was setting the path for the key with the name "(Default)" (which is incorrect, because the name of the default key should be `null`).
This also takes into account the registry key G HUB places that points to `sdk_legacy_led_xXX.dll`, overwriting its value. Should this behaviour be reversible ( adding a unpatch button that would revert the patch? This would need to save to configuration what the original state of the registry was before patching.)

Should fix everyone's problems with GTA 5 and Battlefield I.


simon-wh:
An unpatch button may be necessary if it's not possible to use the Logitech stuff as you'd expect after getting rid of Aurora.

diogotr7:
We'd have to save the previous state somewhere then. This value can not exist at all, have the LGS oath (which works with Aurora regardless), and the g hub path pointing to the legacy dll. A way to revert the dll itself would also make sense if we were to implement this.

simon-wh:
> We'd have to save the previous state somewhere then. This value can not exist at all, have the LGS oath (which works with Aurora regardless), and the g hub path pointing to the legacy dll. A way to revert the dll itself would also make sense if we were to implement this.

I feel like the whole install process for patches needs to be overhauled as it is very messy, maybe it'd be best to leave it to a rewrite of that rather than implementing something hacky now

PR #1581, opened 2019-05-31
Save config on options window closing event

SnakePin:
Now configuration should be saved when options window is closed.

- Changes code to save global configuration on options window of any device in device manager.

**Known issues/To do:**

- None


gitmacer:
I tested it and can confirm, that it is working.

PR #1584, opened 2019-06-01
Update README.md

clrxbl:
Minecraft is in the list twice.


diogotr7:
Thanks for the contribution, but this has already been fixed in the development branch. Development has been a little slow because all the contributors have been busy, but this next release should be relatively soon :)

Issue #546, opened 2017-04-26
Asus aura and corsair link


Bchap9:
You should add support to control rgb headers, graphics cards, motherboards through Asus aura, and cpu coolers through corsair link. Love the program

antonpup:
I have an Asus Aura motherboard, I don't think they've released an SDK for it yet.

As for Corsair, I am pretty sure that "Corsair Link" is just a name, everything RGB goes through CUE, which Aurora already has support for.

BobVul:
There's an SDK in beta now, with release planned at Computex (end of May): https://rog.asus.com/tag/aura-sync-sdk/

caw1:
i´ve got a reply for that issue: i translated it from german : "thank you for your request. Such an enhancement for our graphics or Motherboards is not planned yet. The lighting can only be controlled by the AURA Software.

Marco Janssen
Customer Service Center
ASUS Computer Germany

brainbug89:
@caw1 maybe you should sent Mr. Marco Janssen this link: https://rog.asus.com/articles/coming-soon/the-poseidon-gtx-1080-ti-and-other-new-rog-products-debut-at-jtr-outshine-the-competition/ (last paragraph)

caw1:
Hi,

Finally i've found the Link: 
https://www.asus.com/campaign/aura/global/SDK.html

:) 

If some of you have the Time and the Knowledge to implement that in Aurora Many of us were so happy :)

stbatuhan:
Well there is no support for corsair link actually. I'm using h100i gtx and not compatible of course. Well it's not good at their software too.

Fma965:
Is this something that can be done? i would absolutely love to get my keyboard and mobo in sync :) 

https://github.com/opticalgenesis/AuraCUE 

not sure if the above link is helpful or not?

ersan:
Corsair case lighting is separate from CUE, unfortunately - it all runs through Corsair LINK instead, which is terrible software.  I would absolutely love if Aurora could support Corsair RGB fans and Lighting Nodes but I don't know how possible it is.

There is a third party program called SIV (http://rh-software.com/) that is capable of controlling the LEDs but I don't know to what extent and it is closed source.  Here is a post detailing how to control the LEDs with SIV, if it helps: http://forum.corsair.com/forums/showthread.php?p=892388 (and how to control Corsair DIMM LEDs: http://forum.corsair.com/forums/showthread.php?p=888742)

caw1:
[...]

PR #1602, opened 2019-06-15
Asus aura support

Fluto:
Test Asus build with support for AuraSDK v3

Fixes: #546 

- Add a new device `Asus`, this aims to support all devices that falls under the AuraSDK
- Modifies Strix Flare device to have more LEDs to be accessed

**Known issues/To do:**

- Test Test Test!
- Link to an Asus driver installer


vicdilou:
Hello,
thanks it works great !
But the light is sync with my mouse, can I add another just for aura or sync with another key ?
Thank you

kekkokk:
I've compiled the source of this branch and it loaded the first time.
next step was going to device and select preferred keyboard Strix Flare.
Then when I tried to change the preferred language into Italian it crashed and I'm not able to start it again. 

aurora_debug

Tried also to delete the build folder. can you give me some hint of where it stores the settings so I can clear and start from scratch to be able to reproduce the issue?
Maybe I can help testing with my Claymore.
Also, can you provide me the link of the latest Asus SDK I have to install?

I did a previous integration of my claymore and it works great (only problem I had to compile only in x86 since the dlls were x86)

I know about nothing of windows programming and C#, I'm more a nodejs / c++ dev, but maybe I can help testing

Thanks

kekkokk:
Hey! Asus Claymore owner here!

I get this error: Exception thrown: 'System.InvalidCastException' in Aurora.exe

when doing this cast: 
err_1

kekkokk:
https://github.com/kekkokk/Aurora/tree/asus-aura-support

Started integrating support for ROG Claymore keyboard which seems to be a bit different (Its recognised as a keyboard but if you try to cast to keyboard the program explodes).

I noticed that in async mode, when pressing lots of keys the app crashes saying "the connection with the mainboard was lost" but the keyboard keys are super responsive. (I have a strix mb with rgb ram and strix VGA so I'm testing everything).

In not async mode it worked without a single crash but the keyboard response time is much higher. I can see the visual feedback lag when I press a key.

Let me know what do you think

That first claymore integration is with the numpad on the right only. (since it can be swapped).

kekkokk:
I keep getting AccessViolationException in the foreach Device.Lights
Is this happening to you too?

Fluto:
@kekkokk 
Hey, I don't have too much time to look into the Claymore issue :(. AccessViolationException seems to be happening internally from the API. I'm not sure how to keep it from complaining. I'm waiting for Asus to expose more of the API or release a nice update. (Or till I have more time to investigate)

kekkokk:
Claymore is not a problem anymore, I wrapped it in it's own object. The access violation exceptions occurs only when threaded so It might be a concurrency issue.I think the only solution here is to wait for each device until it has applied the colors and skip the job if another update comes when busy during apply. I'm looking into it. seems to happen also in RGB.net which is the new implementation in the dev branch. At the moment seems that the bottleneck is the Vga (asus strix 1080) which takes up to 57 ms to apply. another interesting thing is that I made a little project which applies the colors every 33 ms (30fps) at every device and it works perfectly.

[Playground](https://github.com/kekkokk/aura_csharp_playground)

The same code applied in aurora is quite laggy, so I need to study how threads are handled inside aurora (again, I'm new in c#)

Hope to find a reliable solution

Fluto:
(I don't have github email enabled for some reason, so I didn't see your reply)
My threaded approach should be waiting per device, but not for all devices. Which is an interesting idea. I have a Pugio Mouse which decides to deadlock internally and make each call 3seconds long, which is why I made the async approach...

PR #1608, opened 2019-06-20
traduction francais

williamwwlr:
y a t'il possibilité de traduire cette application ou de contribuer a le faire


Wibble199:
Going to assume this was supposed to be an issue, not a PR. There are no changes here, just adding dev to master.

Translations for Aurora are coming in a future update that I am working on. I don't want to give a precise time since I don't know how long it will take, but it should be within the next 2/3 months maybe.

Issue #1570, opened 2019-05-22
Overwatch wrapper no longer works since Anniversary update 5/21/19


Caleb10E:
### Expected behavior:

Wrapper layer for Overwatch should work as it does on Razer keyboards after patching Overwatch directory.

### Actual behavior:

After today's update (5/21), the wrapper layer no longer works. It's just solid black.

### How often does this problem occur?

This problem occurs every time.

### Please describe the steps to reproduce this issue:
The issue can be reproduced by doing following steps:

1. Patch Overwatch directory.
2. Open Overwatch and start a game or go to the practice range. None of the keyboard LEDs will light up.

### What is the version of Aurora you are experiencing this issue with?

My version is 0.6.4b.

### Please include any logs, screenshots, or files that are related to this issue:

[Screenshot of Aurora displaying Overwatch wrapper with game open in the training range](https://i.imgur.com/2KBUI85.png)

Ocaeldan:
Hello,
Exact same issue for me (910), event if i delete the 2 files when we patch OW, delete the 2 microsoft Visual C++, take another version of Aurora and delete the Appdata file of Aurora.

Hope my case can help you too !

diogotr7:
I suspect blizzard implemented signature checks on third party dlls as per razers guidelines, so it's very likely that auroras way of interfacing with the game will no longer work :(

torda385:
Is there any way to get the signed dlls from the Razer app? Or maybe just point Aurora to those dlls?

Wibble199:
No, it's not possible to sign a DLL as someone else. Even if it was possible, the signature itself is a checksum of the DLL meaning that if the DLL is different (like ours is) then the signature would be different even if you used the same private key to sign it.

Ange17:
Hello,
This exact issue also happened to me. I tried reinstalling both the Microsoft Visual C + + and the aurora itself also, tried repatching overwatch several times and still did not work. 
Other games works perfectly and it just does it on Overwatch.

Just a question, how can I get the logs for Overwatch? So other people can see it and check what's wrong.

Wibble199:
We know what's wrong with it. It's just not something wrong with Aurora or the wrapper.

Normally, OW ships with the Razer DLL which is a library of code authored by Razer themselves. OW tells this DLL file which keys should be what colour and the DLL then passes this information to the Razer Synapse software which then converts that information into data that can be understood by the keyboard.

[...]

PR #1632, opened 2019-07-12
Fix Razer Chroma support

MoshiMoshi0:
**List any issues that this PR fixes:** #1570

- Adds a new razer layer that enables support for chroma enabled games
- Fixes Dispose not getting called on Layer/LayerHandler when removing the layer or switching the layer type ~~(required to cleanup threads created by the razer layer)~~
- Adds AppDataDirectory and LogsDirectory constants
- Adds RazerManager to Global
- Rearranges "Devices and Wrappers" tab in settings to include razer chroma
- Fixes settings control overflowing outside the bottom of the window
- Fixes NPE hiding original json errors when deserializing profiles
- Fixes default profile being created twice when an exception occurs during profile deserialization

**To Do**:
- [x] ~~Add support for `allow_wrappers_in_background` option~~ **(not needed)**
- [x] Make it easier to install the correct sdk
- [x] Debug logging
- [x] Ability to dump raw data to a file
- [x] Move razer manager to global
- [x] Add razer panel to settings

**Known issues:**
- ~~Needs testing from users~~
- ~~Very rarely requires you to alt-tab to desktop and back to the game to enable lighting (game issue?)~~ **(reported on one game only)**

**Custom build to test:** https://ci.appveyor.com/project/antonpup/aurora/builds/26793401/artifacts
**SDK version support:** v3.5.6

If you have the wrong SDK version you can either use synapse to upgrade or use the download button in the razer layer. If installing manually make sure to stop all razer services first.


Project-Aurora/Project-Aurora/Settings/Layers/RazerLayerHandler.cs
```diff
@@ -0,0 +1,316 @@
...
+            }
+
+            hue *= 60;
+            if (hue < 0.0) hue += 360;
+
+            saturation = (max == 0) ? 0 : 1d - (1d * min / max);
+            value = max / 255d;
+        }
+
+        private Color FromHsv(double hue, double saturation, double value)
```
simon-wh: This function should be moved to ColorUtils

Project-Aurora/Project-Aurora/Settings/Layers/RazerLayerHandler.cs
```diff
@@ -0,0 +1,316 @@
...
+            else if (provider is RzAppListDataProvider appList)
+            {
+                _currentAppExecutable = appList.CurrentAppExecutable;
+                _currentAppPid = appList.CurrentAppPid;
+            }
+        }
+
+        public bool StartDumpingData()
+        {
+            var root = $@"{Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData)}\Aurora\Logs";
```
simon-wh: You should share a constant for this path as you've reused the hardcoded line elsewhere, makes it a bit annoying if you want to easily switch which directory it's using. 
Side note: We really should have a global constant for the settings directory as if we ever want to change that, we'd have to change a lot of code

Project-Aurora/Project-Aurora/Settings/Layers/RazerLayerHandler.cs
simon-wh: Same with the other generic color manipulation functions you've got in there

MoshiMoshi0: WrapperLightsLayerHandler also has functions for hsv <-> rgb, should i remove them from there too in this pr? And also use ColorUtils?

MoshiMoshi0: Also by other color manipulation functions you mean those in `PostProcessColor`? I dont know if those are that generic since brightness and saturation boosting can be done in different ways. I did it so boost value of 0 equals 0% boost and boost value of 1 equals inf% boost, i dont know if that will be self explanatory if I put them in ColorUtils.

Project-Aurora/Project-Aurora/Settings/Layers/RazerLayerHandler.cs
MoshiMoshi0: Added AppDataDirectory and LogsDirectory constants and replaced in all places it was used.

Project-Aurora/Project-Aurora/Settings/Layers/RazerLayerHandler.cs
simon-wh: Yeah, throw them into ColorUtils as well so they're all shared. For the PostProcessColor you could have that as a particular way of applying the boosting in ColorUtils, other ones could be added under a different name if needed in ColorUtils.

MoshiMoshi0: Should be done

diogotr7:
The default profiles for at least overwatch should be changed to have a chroma layer instead of the old wrapper one.

zejjnt:
Unfortunately the Corsair RGB fans stopped working in 0.6.4b again :/

MoshiMoshi0:
I did not change anything with corsair so it must have been broken by some other pr merged to dev branch.
You should create a new issue about this, or find existing one.

PR #1633, opened 2019-07-14
Added Coordinates ambilight type and fixed a couple of bugs

diogotr7:

* Renamed the "Everything" ambilight type to "Entire monitor" as that's what it actually does now.
* Fixed the foreground and specific process options that were not copping correctly
* Added a Coordinates option, which crops the screenshot to the given values. might be useful to get health bars and stuff like that fron the screen.
* Improved the screenshot quality to hopefully make this Coordinates option more useful, as cropping a screenshot that's already been resized to the canvas size doesn't look very good.
* Reduced the amount of times the average color gets calculated. For some reason it was getting computed more than needed before.

**Questions/Todos**
* ~~Better UI/UX for the coordinates menu?~~
   * ~~Grouping them in boxes might be a good idea~~
* ~~Maybe the screenshot can be shrunk even more without noticeable loss in details(which aren't that noticeablew on a keyboard to begin with)?~~


Project-Aurora/Project-Aurora/Settings/Layers/AmbilightLayerHandler.cs
```diff
@@ -28,17 +28,17 @@ public enum AmbilightType
...
+        [Description("Entire Monitor")]
+        EntireMonitor = 0,
 
         [Description("Foreground Application")]
-        ForegroundApp = 2,
+        ForegroundApp,
 
         [Description("Specific Process")]
-        SpecificProcess = 3
+        SpecificProcess,
```
simon-wh: I'd be wary about reordering this enum, as it means that people's existing settings will change about when they use the new version

Project-Aurora/Project-Aurora/Settings/Layers/AmbilightLayerHandler.cs
```diff
@@ -87,6 +87,26 @@ public class AmbilightLayerHandlerProperties : LayerHandlerProperties2Color<Ambi
         [JsonIgnore]
         public String SpecificProcess { get { return Logic._SpecificProcess ?? _SpecificProcess ?? String.Empty; } }
 
+        public int? _CoordinateX { get; set; }
```
simon-wh: You could probably replace all these coordinate variables with the Rectangle type

Project-Aurora/Project-Aurora/Settings/Layers/AmbilightLayerHandler.cs
diogotr7: Should i set Coordinates to be '0' instead? The Everything option doesn't exist anymore.

diogotr7: I decided to put the coordinates at 0. Would another way be better?

simon-wh: Hmmm, I'm not sure. You could potentially set the default coords to -1 for all and have that result in it taking the entire screenshot so the behaviour is the same? How would you do all screens with the changes you've made?

diogotr7: Setting the defaults to -1 won't work, as the coordinates are relative to your primary monitor. Negative ones are valid if you have monitors above or to the left of the main one. I tried to make this clearer by disabling the "Output ID" textbox whenever the user as any option other than "Entire Monitor" selected. Does that make sense?

lucasvanhalst:
Nice, but it seems the information regarding updates per second is shown at the wrong position

diogotr7:
True, I moved the comboboxes and forgot the warnings. Will fix.

simon-wh:
Looks really good, thanks!

PR #1638, opened 2019-07-21
Fixes dota2 ability layer not handling seasonal abilities

diogotr7:
New High five ability and the seasonal abilites break the current ability layer, causing the whole profile to be black.

This skips any ability that isn't relevant (no cooldown).

### DIFF
--- a/Project-Aurora/Project-Aurora/Profiles/Dota 2/Layers/Dota2AbilityLayerHandler.cs
+++ b/Project-Aurora/Project-Aurora/Profiles/Dota 2/Layers/Dota2AbilityLayerHandler.cs
@@ -71,8 +71,11 @@ namespace Aurora.Profiles.Dota_2.Layers
                     for (int index = 0; index < dota2state.Abilities.Count; index++)
                     {
                         Ability ability = dota2state.Abilities[index];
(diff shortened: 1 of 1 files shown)


Project-Aurora/Project-Aurora/Profiles/Dota 2/Layers/Dota2AbilityLayerHandler.cs
```diff
@@ -71,8 +71,11 @@ public override EffectLayer Render(IGameState state)
                     for (int index = 0; index < dota2state.Abilities.Count; index++)
                     {
                         Ability ability = dota2state.Abilities[index];
-                        if (ability.Name.Contains("seasonal"))
+                        if (ability.Name.Contains("seasonal") || ability.Name.Contains("high_five"))
+                        {
+                            index++;
```
sreichholf: This is a double increment, i guess that wasn't intended?

diogotr7: Correct, i shoudn't code late at night :D.

Shouldn't matter anyway because the abilties we want to skip are at the end anyway.

Thanks, ill change that.

PR #1640, opened 2019-07-23
Fixed crash when reading locked Witcher 3 config file

diogotr7:


Project-Aurora/Project-Aurora/Profiles/Witcher3/GameEvent_Witcher3.cs
```diff
@@ -66,79 +63,64 @@ public override void ResetGameState()
...
         public override void UpdateLights(EffectFrame frame)
         {
             Queue<EffectLayer> layers = new Queue<EffectLayer>();
 
             if (File.Exists(dataPath))
             {
-                if(configContent != null)
+                if (configContent != null)
                 {
-                    var signRes = _configRegex.Match(configContent);
```
Wibble199: Possibly want to be careful with using a regex each update as they may not always be as fast as other methods. While not being able to see what a config looks like, I think for this instance you could probably use `String.IndexOf` method to achieve the same.

Though if performance is fine currently, then this is probably just nit-picking.

Project-Aurora/Project-Aurora/Profiles/Witcher3/GameEvent_Witcher3.cs
```diff
@@ -66,79 +63,64 @@ public override void ResetGameState()
...
-            get { return this.Application.Settings.IsEnabled && isInitialized; }
-        }
-
         public override void UpdateLights(EffectFrame frame)
         {
             Queue<EffectLayer> layers = new Queue<EffectLayer>();
 
             if (File.Exists(dataPath))
             {
-                if(configContent != null)
```
Wibble199: How often does the config change? If it is not changing as often as each frame ticks, this code here will be running unnessecarily.

You could define a boolean field for whether the game state is 'dirty'. This could be set to true every time the config file is re-read and then only run this part of the code if `configContent != null && isStateDirty`?

Obviously if the config is being written very regularlly, this wouln't achieve anything. And again, if performance is okay at the moment this is maybe a non-issue.

diogotr7: The mod writes to te config every 30 player ticks. I'm not sure what exactly this means but seems to be less often than an aurora frame tick. I implemented this with the latest commit.

Project-Aurora/Project-Aurora/Profiles/Witcher3/GameEvent_Witcher3.cs
diogotr7: IndexOf is indeed useful for this. I rewrote most of the parsing in the latest commit and it's not using regular expressions anymore. Should also be less repetitive and more readable.

As for performance, i played a couple of minutes and it seemed a little better than before but it's probably placebo. In any case it's an improvement as it doesnt have to do as much processing each tick.

PR #1664, opened 2019-08-20
Added OnActivate and OnDeactivate methods to ILightEvent

DrParanoia:
Added OnActivate and OnDeactivate methods which should be called when the profile becomes active/inactive

`ILightEvent` now has two more methods - `OnActivate` and `OnDeactivate` which are called when the Application is being brought to foreground or vice versa.

**Known issues/To do:**
Needs to be tested if new methods are called in all cases


DrParanoia:
Never mind, will do some more testing and re-submit.

PR #1666, opened 2019-08-21
OnStart and OnStop callback methods for ILightEvent

DrParanoia:
- adds and manages `OnStart` and `OnStop` callbacks in `ILightEvent`

**Why this is needed:**
Currently, if a game profile uses for example a `FileSystemWatcher` (like in `Witcher 3` profile), the watcher is created in the constructor of the `Event`. This means, that this `FileSystemWatcher` is active even if the profile is not in use. Multiply this by the number of games that need this, and you may end up with A LOT of active event listeners which take up system resources.

The situation get's worse when there is a need for more CPU intensive tasks to be performed by the profile. For example, in order to create acceptable user experience for `Elite: Dangerous` `FileSystemWatcher` can't be used, since it introduces a delay for up to 2 seconds. Instead, a file read loop with a 5ms delay needs to be used which needs to be disabled if the profile is not active to save resources.

**How this PR solves the issue:**
`LightStateManager`s `Update` method updates only active `ILightEvent`s. This means that if an `ILightEvent` was not updated by the end of the `Update` method, it can be considered inactive.

The proposed code keeps track of which `ILightEvent`s were updated and calls the `OnStart` method on each of the `ILightEvent` (if it wasn't called already). `ILightEvent`s which had their `OnStart` method called are stored in a `StartedEvents` list.

If by the end of the `Update` method a certain `ILightEvent` was not updated but was previously started (had it's `OnStart` method called), this event is removed from the `StartedEvents` and the `OnStop` method is called.

**TLDR:**
This PR allows to save system resources by letting game profiles know when they are in use and when they become inactive (via the `OnStart` and `OnStop` methods) so according resource saving actions can be performed (disable file reading loops, clean memory, unregister event listeners, etc.).


simon-wh:
Looks great, thank you!

PR #1680, opened 2019-08-31
Roccat Tyon support

Popat0:

- Added support for the Roccat Tyon mouse using Unified HID.

**Known issues/To do:**

- Both light zones have to be piloted together otherwise it breaks the mouse ability to switch profiles for some odd reason.


Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            return ctrl_device.WriteFeatureData(initPacket);
+        }
+
+        static bool WaitCtrlDevice()
+        {
+            for (int i = 1; i < 100; i++) // If still fails after 100 tries then timeout
+            {
+                // 150ms is the magic number here, should suffice on first try.
+                Thread.Sleep(150);
```
simon-wh: I'm a bit concerned about this loop and the Thread sleeping, it could result in a significant delay to other devices if it ends up going through this loop many times. 

Side note: Defo need to make the Device connect functions async

Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
```diff
@@ -0,0 +1,153 @@
...
+                hwmap[1] = green;
+                hwmap[2] = blue;
+                hwmap[3] = 0x00;
+                hwmap[4] = 0x00;
+                hwmap[5] = red;
+                hwmap[6] = green;
+                hwmap[7] = blue;
+                hwmap[8] = 0x00;
+                hwmap[9] = 0x80;
+                hwmap[10] = 0x80;
```
simon-wh: Could save some lines and make it look nicer by putting these in the initialisation of the byte array rather than assigning the values to each index

Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                byte[] workbuf = new byte[30];
+                Array.Copy(controlPacket, 0, workbuf, 0, 19);
+                Array.Copy(hwmap, 0, workbuf, 19, 11);
+
+                unsafe
+                {
+                    fixed (byte* workbufPointer = workbuf)
+                    {
+                        if (ctrl_device.WriteFeatureData(workbuf) != true)
```
simon-wh: bool != bool?
`if (!ctrl_device.WriteFeatureData(workbuf))`

simon-wh (changes requested):

Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
Popat0: Fixed that up, thanks.

Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
Popat0: As said on discord, that code was on the other Roccat device.
I removed the thread sleeping and reduced the loop to 3 (removing the loop completely made it so the mouse was always stuck on the init colors)

Project-Aurora/Project-Aurora/Devices/UnifiedHID/RoccatTyon.cs
Popat0: Done did, thanks.

simon-wh (approved):

simon-wh:
Great work, thanks for the contribution!

StrykeDev:
It stopped working after a few days (of fresh windows installation).
When starting the service it set the device to white but it doesn't update the color.

PR #1713, opened 2019-09-16
Enum Comparison Evaluatable

Wibble199:
This PR adds an `IEvaluatable<bool>` that determines if the specified enum game state parameter matches the given value.

image

Once the user has chosen a game state, the available enum values then populate the second dropdown. Note that if the enum value has a description (using the `DescriptionAttribute`), this will be the value displayed in the dropdown.

### DIFF
--- a/Project-Aurora/Project-Aurora/Profiles/LightingStateManager.cs
+++ b/Project-Aurora/Project-Aurora/Profiles/LightingStateManager.cs
@@ -89,6 +89,7 @@ namespace Aurora.Profiles
 
         private ActiveProcessMonitor processMonitor;
         private RunningProcessMonitor runningProcessMonitor;
+        public RunningProcessMonitor RunningProcessMonitor => runningProcessMonitor;
 
         public LightingStateManager()
         {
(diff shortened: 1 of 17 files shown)


Wibble199:
Futher added a few more evaluatables.

### Process Running
Outputs a boolean indicating if the given process is running or not. Uses the detection method added for the overlay system and is more performant that using `Process.GetProcessesByName`.

### Away Timer
Outputs a true boolean after the user has not pressed a keyboard or mouse button in a configurable amount of time. Once a button has been pressed, outputs false and restarts the timer.

### Away Time
Similar to the away timer but instead returns a number in a configurable unit (milliseconds, seconds, minutes, hours) since a button was pressed.

### Numeric Change Detector
Outputs a true signal when the attached numeric evaluatable's value changes. Can be configured to listen to only increases, only decreases and only changes of a certain amount. Will only output a true for the single evaluatable tick that the change occurs, so it may be useful to use in conjunction with the...

### True Extender
Extends a true evaluatable to be of a certain, configurable duration.

PR #1742, opened 2019-10-04
fix: Equalizer sending exception to log when no area is drawn

srodriguez1850:
If the equalizer layer is set to use assigned keys (instead of freestyle), but there's no keys, it'll throw DivideByZero exception because the rectangle has a width of 0.

Minor issue (it writes to the log every frame, no stacktrace), but also because it can accidentally slip to the user when selecting assigned/freestyle, and not assign anything.


diogotr7:
Thanks!

PR #1743, opened 2019-10-04
Added support for Divinity2: Original Sin

N-Schaef:

- Add support for Divinity 2: Original Sin

**Known issues/To do:**


N-Schaef:
Divinity 2 Original Sin uses Razer Chroma, I thereby copied and adapted the code from the DOOM profile.

This integration supports the classic and definitive edition.

diogotr7:
I'm wondering what to do about adding in generic wrapper game profiles.

On one hand it makes it easier for users to know which games are supported, but it also clutters the interface and codebase a lot with the same code over and over again. I was thinking of maybe doing the same I've done to the devices and split profiles up into dlls.

Would like opinions on this.

diogotr7:
Upon giving this some more thought, i dont think it makes sense to keep adding simple integrations like this. In the future i'd like to make integrations and devices more modular, ad that would allow for adding games like this faster and easier.

For now, I'll try to update the documentation that adding a razer layer to the desktop profile would make all razer games work corectly (assuming compatible installation of the razer sdk).

Thanks for contributing, though! :)

Issue #1647, opened 2019-07-28
Native Philips Hue Support


wolfcomp:
For people that have wanted a actual native Philips Hue support i'm working on getting it implemented in my fork and later making a PR for it.

Right away it will only react to one specific spot witch will be on Peripheral Logo some point later down the line i'm going to make a customizable placement map for where it should interact.

PR #1757, opened 2019-10-16
Native Philips Hue Suport

wolfcomp:
Closes: #1647 

### Adds
Philips Hue syncing

### Known issues that are impossible to fix in Aurora
If you have a event set in the hue app on the phone it will try to activate that event on the lights but Aurora takes over the light syncing again the next time it sends to the lights making a slight blink effect.

### Images
931961e300
a215428906

Note all the previous config will work with this new version as this is only for splitting the pull requests. And to remove my custom funky mess when it comes to how i have to build a custom version for it to work fully on my setup.


Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
```diff
@@ -97,116 +94,36 @@ public void SetVariable(object newvalue)
...
-                    Value = b.Value;
-                    break;
-                case BindableColor b:
-                    Value = b.Value;
-                    break;
-                case Bindable<string> b:
-                    Value = b.Value;
-                    break;
+                Value = (Bindable as Bindable<object>).Value;
+                (Bindable as Bindable<object>).ValueChanged += _ => Value = (Bindable as Bindable<object>).Value;
```
Wibble199: As before, you cannot cast a generic type to a different generic type even if the type parameter of one extends the type param of the other.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using Aurora.Utils;
+
 using Newtonsoft.Json;
+
 using System;
```
Wibble199: What's with these sort of unnecessary changes? We've not got any whitespace in the imports in other files.

Project-Aurora/Project-Aurora/Devices/DeviceManager.cs
```diff
@@ -96,23 +96,24 @@ public int RetryAttempts
...
+            devices.Add(new DeviceContainer(new Devices.UnifiedHID.UnifiedHIDDevice()));     // UnifiedHID
+            devices.Add(new DeviceContainer(new Devices.Wooting.WootingDevice()));           // Wooting
+            devices.Add(new DeviceContainer(new Devices.Creative.SoundBlasterXDevice()));    // SoundBlasterX
+            devices.Add(new DeviceContainer(new Devices.LightFX.LightFxDevice()));           // Alienware
+            devices.Add(new DeviceContainer(new Devices.Dualshock.DualshockDevice()));       // DualShock 4
+            devices.Add(new DeviceContainer(new Devices.Drevo.DrevoDevice()));               // Drevo
+            devices.Add(new DeviceContainer(new Devices.YeeLight.YeeLightDevice()));         // YeeLight
+            devices.Add(new DeviceContainer(new Devices.Asus.AsusDevice()));                 // Asus
+            devices.Add(new DeviceContainer(new Devices.NZXT.NZXTDevice()));                 // NZXT
+            devices.Add(new DeviceContainer(new Devices.Hue.HueDevice()));                   // Philips Hue
```
Wibble199: There's really no need for these sorts of very minor cosmetic changes, especially when it falls outside of the scope of your PR.

Project-Aurora/Project-Aurora/Settings/Bindables/Bindable.cs
```diff
@@ -0,0 +1,231 @@
...
+using Aurora.Utils;
+
+using Corale.Colore.Annotations;
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Aurora.Settings.Bindables
+{
+    public class Bindable<T> : IBindable<T>, ISerializableBindable
```
Wibble199: What sort of issues are these "Bindables" solving that aren't already solved by the existing VarRegistry. While the VarRegistry is by no stretch of the imagination a great example of config management, it has been doing the job for other devices for a long time. Whilst we are not against changing the VarRegistry by any means, having spoken to the other contributors, we currently do not see any advantage of using this over the existing method and we certainly don't want to have multiple config management systems doing effectively the same thing.

Project-Aurora/Project-Aurora/Settings/Bindables/BindableNumber.cs
```diff
@@ -0,0 +1,428 @@
...
+                    break;
+
+                case TypeCode.Double:
+                    var doubleBindable = this as BindableNumber<double>;
+                    if (doubleBindable == null) throw new ArgumentNullException(nameof(doubleBindable), $"Generic type {typeof(T)} does not match actual bindable type {GetType()}.");
+
+                    doubleBindable.Value += Convert.ToDouble(val);
+                    break;
+            }
+        }
```
Wibble199: Is there any reason to use switch trees like this? Why not get the type of `this` and then from there get the generic parameter that has been defined on it and use `Convert.ChangeType` and pass the generic parameter type in? It would mean you could remove a lot of these big chunks of code.

Project-Aurora/Project-Aurora/Settings/Bindables/ExtraBindables.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    var k = f.Split('=');
+                    return new KeyValuePair<string, string>(k[0], k[1]);
+                }).ToDictionary(f => f.Key, f => f.Value);
+                return new RealColor(Color.FromArgb(Convert.ToInt32(h["A"]), Convert.ToInt32(h["R"]), Convert.ToInt32(h["G"]), Convert.ToInt32(h["B"])));
+            }
+            set => base.Default = value.ToString();
+        }
+    }
+}
```
Wibble199: Why are you storing this as a string? It means you're having to parse and stringify the value whenever it's read and written.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
```diff
@@ -20,21 +24,66 @@ public class VariableRegistryItem
...
+                    break;
+                case Bindable<string> b:
+                    Value = b.Value;
+                    Default = b.Default;
+                    break;
+            }
+            Title = title;
+            Remark = remark;
+            Flags = flags;
         }
```
Wibble199: What if I want to add a new bindable type? Do I have to add it in here too? This has the potential to become another massive chunk of code. And what about the other BindableNumbers you have, such as short?

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
```diff
@@ -231,13 +288,12 @@ public void RemoveVariable(string name)
...
+            string str = JsonConvert.SerializeObject(this, Formatting.None, new JsonSerializerSettings {TypeNameHandling = TypeNameHandling.All, Binder = Aurora.Utils.JSONUtils.SerializationBinder});
             return JsonConvert.DeserializeObject(
-                    str,
-                    this.GetType(),
-                    new JsonSerializerSettings { ObjectCreationHandling = ObjectCreationHandling.Replace, TypeNameHandling = TypeNameHandling.All, Binder = Aurora.Utils.JSONUtils.SerializationBinder }
-                    );
+                str,
+                this.GetType(),
+                new JsonSerializerSettings {ObjectCreationHandling = ObjectCreationHandling.Replace, TypeNameHandling = TypeNameHandling.All, Binder = Aurora.Utils.JSONUtils.SerializationBinder}
+            );
```
Wibble199: You really need to stop committing code that changes styling changes like this. It's unnecessary, and the project's general style is to have spaces around the braces. Even if your IDE is doing these changes automatically, you should be looking at what you are committing before doing so intead of just adding all changes.

Wibble199 (commented):
Not looked over all the code but these are some of the things that stood out at problematic.

Project-Aurora/Project-Aurora/Settings/Bindables/ExtraBindables.cs
wolfcomp: The reason comes down to how the color is stored in run time where i need to export it reliably to the config witch is a JSON in a seperate file and how the config is loaded as this extends a Bindable<string> and not a color definition.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
wolfcomp: Honestly you guys should have looked at supplying a ReSharper settings file that i could use so i don't have these issues in the repo.

Project-Aurora/Project-Aurora/Settings/Bindables/Bindable.cs
wolfcomp: When i wrote the hue code and i tried to actualy get the values back from the VarRegistry they were the default values and never returned the actual changed values and when i implemented this it fixed the issue.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
wolfcomp: https://github.com/antonpup/Aurora/pull/1757#discussion_r336581440

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
```diff
@@ -97,116 +94,36 @@ public void SetVariable(object newvalue)
...
-                            b.Value = (long)newvalue;
-                            break;
-                        case BindableColor b:
-                            b.Value = (RealColor)newvalue;
-                            break;
-                        case Bindable<string> b:
-                            b.Value = (string)newvalue;
-                            break;
-                    }
+                    (Bindable as Bindable<object>).Value = newvalue;
```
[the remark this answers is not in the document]
wolfcomp: Yes i tested this and all the settings was being saved and stored properly in the code.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
wolfcomp: https://github.com/antonpup/Aurora/pull/1757#discussion_r336582007

Project-Aurora/Project-Aurora/Devices/DeviceManager.cs
wolfcomp: Honestly those comments don't need to be there _as mentioned in the discord_ since we can clearly see what is what

Project-Aurora/Project-Aurora/Settings/Bindables/BindableNumber.cs
wolfcomp: I would need to test that out before i can comment further than this is a solution i found from another repo that is MIT licenced.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
wolfcomp: A tiny little miss yesterday when i changed most of the switches over to Bindable<object> as i was really tired when i made that change.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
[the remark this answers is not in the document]
Wibble199: But why chose to create a whole new system over attempting to resolve the issue in the existing code?

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
Wibble199: I don't think it's really on us to have to provide a file for a 3rd party proprietry software.

Project-Aurora/Project-Aurora/Settings/Bindables/Bindable.cs
Wibble199: But why did you choose to implement a new system just for the Hue over attempting to fix the code that is already in place in many other places?

wolfcomp: As when i break pointed the code it wasn't even changed in the VarRegistry meaning it was something in WPF that most likely caused it not to be set.

Project-Aurora/Project-Aurora/Settings/VariableRegistry.cs
wolfcomp: ReSharper isn't a proprietry software and is actual being used quite the bit in actual companies. There exists a 30 day trail for it meaning it can be configured up to how the code should be removing these issues later down the line.

Wibble199: Right, but if you wanna keep using it beyond the 30 days you need to pay for it. Therefore, proprietry.

Either way, if it is or isn't on us to provide a config file, it doesn't mean it's open to random style changes that add nothing.

RaberShef:
This merge is waiting since last year can someone merge it already?

wtf911:
When is this going to be added? Would love this to work with my Hue lights please!!!!!!!!!!!!!!

wolfcomp:
Will close this for now as im to busy to keep up with the code changes and other stuff going on.

PR #1767, opened 2019-10-21
Support for Rosewill Neon RGB k85

Xereco:
If possible can I add support for my Rosewill keyboard myself?

- 
- 
- 

**Known issues/To do:**

- 
- 
-


Wibble199:
I think Rosewill products has been requested before, however there is no SDK available for us to use.

To be able to add support for a particular device (keyboard, mouse, etc.), we need to be able to access an SDK. A Software Development Kit (SDK) is a code library that allows manufacturers to provide a way of controlling things with external programs, like Aurora. It has various procedures programmed into it that developers can use to send our own colour data to the keyboards or other devices. The SDK handles things such as making sure the data is in the right format, establishing a connection with the keyboard and anything else that needs to be done to make it work.

---

Closing this PR since it should be an issue.

PR #1774, opened 2019-10-28
Reworked ambilight layer and added a vertical flip option

### LINKED ISSUE #442: Ambilight vertical flip?
Would it be possible to flip ambilight vertically? It makes more sense to me for it to look reflective.

diogotr7:
Fixes #442

* Adds an option that flips the image vertically. This improves the illusion that th keyboard is reflecting the screen.
* Changed the way the cropping is done. Before, the image was cropped differently for each screen capture mode. Now, the target screen region is calculated from the options and then applied to the layer. Should do the same thing but is a bit cleaner imo. Code review appreciated.
* Adds the ability to display the ambilight image in a keyboard region instead of covering the whole canvas.
* **Reworks the 2 implementations of ambilight into an interface + 2 classes for better clarity.**

**Known issues/To do:**
* The access lost exception is still not fixed, not sure what causes this exactly,
* Only one DX desktop duplicator layer should be active at any time. There needs to be a system trhat prevents then user from turning on the "experimental mode" if **any other layer on the aplication** has it turned on already.


gitmacer:
Instead of adding the flip option just to the ambilight layer. I would suggest to add a mirror vertically and horizontally option to the layer options here:
grafik

gitmacer:
- In my opinion the Monitor selection should start at 1 instead of 0
- Also Wibbles process picker should be used?

PR #1777, opened 2019-11-01
Added device discover feature for YeeLight and lowered the color change delay

SnakePin:

- Adds device discover feature for YeeLight.
- Lowers the color change delay from 100 to 35.
- Updated YeeLight library version to 0.0.6

**Known issues/To do:**

- Nothing.


Project-Aurora/Project-Aurora/Devices/YeeLight/YeeLightDevice.cs
```diff
@@ -118,27 +116,43 @@ private int GetFreeTCPPort()
...
                 }
+                else
+                {
+                    throw new Exception("IP address is set to 0.0.0.0 and no devices has been located.");
+                }
+            }
+            //If it isn't then set the IP to the provided IP address
+            else
+            {
+                light.SetLightIPAddressAndPort(lightIP, YeeLightAPI.YeeLightConstants.Constants.DefaultCommandPort);
```
xQwexx: It's a minor thing, you uses the api constant in below and below the local constant which a little bit confusing for me, because both value is the same. so I rather declare the local cont as private const int lightListenPort = YeeLightAPI.YeeLightConstants.Constants.DefaultCommandPort; and use this local variable after it.

Project-Aurora/Project-Aurora/Devices/YeeLight/YeeLightDevice.cs
```diff
@@ -118,27 +116,43 @@ private int GetFreeTCPPort()
...
-            catch (Exception exc)
+
+            IPAddress localIP;
+            int localListenPort = GetFreeTCPPort(); // This can be any free port
+
+            using (Socket socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, 0))
             {
-                Global.logger.Error($"Device {devicename} encountered an error during Connecting. Exception: {exc}");
-                isConnected = false;
+                socket.Connect(lightIP, lightListenPort);
```
xQwexx: If the IP is 0.0.0.0, so discovery feature is on,  the lightIP is not the discovered device IP. This case it's always 0.0.0.0, I am not sure it is going to cause any problem.

Project-Aurora/Project-Aurora/Devices/YeeLight/YeeLightDevice.cs
```diff
@@ -79,8 +79,6 @@ public bool IsPeripheralConnected()
 
         public bool Reconnect()
         {
-            light.CloseConnection();
-
```
xQwexx: I don't understand why "light.CloseConnection();" is not needed, because if light is connected then nothing going to close it. So the isConnected = false;  value don't going to change in Connect(), because light.IsConnected() is true.

Project-Aurora/Project-Aurora/Devices/YeeLight/YeeLightDevice.cs
SnakePin: I should have named the local variable better, basically there is a local listen port needed for switching the light to music mode which allows us to change color without getting rate limited, and there is a port on the light which we can connect to but be rate limited, we connect to that command port on the light and tell it that our local listen port is [something] and then light opens a connection to our device. This code tries to find a free local port and sends it to the light's command port.

Project-Aurora/Project-Aurora/Devices/YeeLight/YeeLightDevice.cs
SnakePin: Thanks, good finding, I'll try to fix this, maybe even that's the reason why device discover wasn't working correctly.

Hukuma1:
Is there a way to test this now and not wait for the merge? I got the nightly Aurora 0.7.2+809 beta. Not sure where to place YeeLightDevice.cs? Guessing it's not used because the delay is still at 100 and not the new 35 when checking options.

Hukuma1:
```
[ERROR] 08:58:37: Device YeeLight encountered an error during connecting. Exception: System.TimeoutException: The operation has timed out.
   at YeeLightAPI.YeeLightDevice.SetMusicMode(IPAddress localIP, UInt16 localPort, Boolean state)
   at Aurora.Devices.YeeLight.YeeLightDevice.Connect(IPAddress lightIP, DoWorkEventArgs token) in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 162
   at Aurora.Devices.YeeLight.YeeLightDevice.Initialize() in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 48
```

Getting this error. That is not my path to `YeeLightDevice.cs`, so not sure if it's because of that. Do any ports have to be opened on router btw?

SnakePin:
> 
> 
> ```
> [ERROR] 08:58:37: Device YeeLight encountered an error during connecting. Exception: System.TimeoutException: The operation has timed out.
>    at YeeLightAPI.YeeLightDevice.SetMusicMode(IPAddress localIP, UInt16 localPort, Boolean state)
>    at Aurora.Devices.YeeLight.YeeLightDevice.Connect(IPAddress lightIP, DoWorkEventArgs token) in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 162
>    at Aurora.Devices.YeeLight.YeeLightDevice.Initialize() in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 48
> ```
> 
> Getting this error. That is not my path to `YeeLightDevice.cs`, so not sure if it's because of that. Do any ports have to be opened on router btw?

You need to make sure Aurora is allowed for both outbound and inbound networking in Windows Firewall, follow https://support.microsoft.com/en-us/help/4558235/windows-10-allow-an-app-through-microsoft-defender-firewall

Hukuma1:
> > ```
> > [ERROR] 08:58:37: Device YeeLight encountered an error during connecting. Exception: System.TimeoutException: The operation has timed out.
> >    at YeeLightAPI.YeeLightDevice.SetMusicMode(IPAddress localIP, UInt16 localPort, Boolean state)
> >    at Aurora.Devices.YeeLight.YeeLightDevice.Connect(IPAddress lightIP, DoWorkEventArgs token) in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 162
> >    at Aurora.Devices.YeeLight.YeeLightDevice.Initialize() in C:\projects\aurora\Project-Aurora\Project-Aurora\Devices\YeeLight\YeeLightDevice.cs:line 48
> > ```
> > 
> > 
> > Getting this error. That is not my path to `YeeLightDevice.cs`, so not sure if it's because of that. Do any ports have to be opened on router btw?
> 
> You need to make sure Aurora is allowed for both outbound and inbound networking in Windows Firewall, follow https://support.microsoft.com/en-us/help/4558235/windows-10-allow-an-app-through-microsoft-defender-firewall

That worked. Thanks! What's weird is I did not have this rule in before and it seemed to work in an earlier version. Maybe updating to new Win10 v2004 reset something.

Hukuma1:
If selecting Away Effects in Aurora and doing Blackout (turns everything black/off), Yeelight stays at same full brightness and last color (e.g. red). Is there a way to dim it at least? I know if you turn it off it loses connection. But at least maybe set brightness to 0?

SnakePin:
It's not possible to set brightness to 0 and even 1 is bright enough to light a room.

Hukuma1:
Which Yeelight bulb and firmware are you using? This has been definitely addressed I think. Can show you photos of my 1S bulbs (YLDP13YL) with minimum brightness vs maximum brightness. Or you mean those settings aren't exposed in API?

Hukuma1:
If you could add a brightness slider into your plugin (global brightness for all effects is okay) I could show you difference on 1S bulbs.

SnakePin:
What I'm saying is that API doesn't allow brightness value 0, it's even specified in the API documents.

Hukuma1:
Ah, yes I see that.

```
Method:set_brightUsage:This method is used to change the brightness of a smart LED.Parameters:3."brightness" is the target brightness. The type is integer and rangesfrom 1 to 100. The brightness is a percentage instead of a absolute value. 100 meansmaximum brightness while 1 means the minimum brightness."effect": Refer to "set_ct_abx" method."duration": Refer to "set_ct_abx" method.Request Example: {"id":1,"method":"set_bright","params":[50, "smooth", 500]}Response Example: {"id":1, "result":["ok"]}NOTE:Only accepted if the smart LED is currently in "on" state.
```

Could you please try setting of 1?

SnakePin:
It won't be any useful, maybe setting to white when the color is black could be done though, as white is the default color.

Hukuma1:
Is there a way to turn down brightness on the specific key I got yeelight connected to at least then? e.g. all keys 100% brightness except ESC key at 25% brightness

PR #1789, opened 2019-11-10
Added vulcan implementation using a nuget package

diogotr7:
Removes the unified HID implementation and uses a nuget package instead


Project-Aurora/Project-Aurora/Project-Aurora.csproj
```diff
@@ -102,6 +102,9 @@
     <Reference Include="HidLibrary, Version=3.2.46.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\hidlibrary.3.2.46.0\lib\HidLibrary.dll</HintPath>
     </Reference>
+    <Reference Include="HidSharp, Version=2.1.0.0, Culture=neutral, processorArchitecture=MSIL">
```
simon-wh: Howcome we have a dependency to HidSharp now?

diogotr7: That's what Vulcan.net uses to interface with the keyboard. Hidlibrary is too slow. Snakepin suggested we transition unifiedhid over to hidsharp as well

simon-wh: Correct me if I'm wrong, but shouldn't the fact that it's a dependency of Vulcan.NET mean that the dependency is on Vulcan rather than Aurora?

PR #1796, opened 2019-11-12
Fix various bugs with Minecraft - Fixes #1794

### LINKED ISSUE #1794: Minecraft Profile uses outdated Conditional Layers
The water, sneaking, horse and player control keys use conditional layers (which are deprecrated).

stuntguy3000:
**List any issues that this PR fixes:** Fixes #1794

- Changes Conditional Layers to Solid Colo(u)r Layers with Override Logic
- Change "Keys" to "Controls" for clarity
- Add more conditional logic to hide special effect layers when player is not in game

**Known issues/To do:**

- Fully tested and working.


stuntguy3000:
The only issue I have with this is that it doesn't update existing profiles and only changes new installations. Is it possible to force a profile reset, like done for Rocket League's latest update?

diogotr7:
I think in most cases using overrides is a good idea but honestly i like conditional layers because if a user doesn't know about overrides, it's impossible for them to know why a solid color layer isn't showing up. A conditional layer is easier to understand.

The rest of the changes make sense.

PR #1799, opened 2019-11-12
Update Control_LoL.xaml with new file path in description

aurorasmiles:
Just updates the information in the Leage of Legends Profile overview where the LeagueofLegends.exe file is usually found.


simon-wh:
PR's should be to dev not master

PR #1804, opened 2019-11-13
Custom led mapping for Asus and Corsair

diogotr7:
Adds 2 text files with mappings between aurora's DeviceKeys and RGB.NET's LedId. Allows a user to edit a text file to, for example, use tha additional lights on the k95 plat layout to control case lights individually from the peripheral logo.

Could use some advice in how to do this better as currently the code is a bit weird.


diogotr7:
This will not be merged. If people find this useful, I'll add the feature to the corsair dll in my `feature/device-dll` branch

Issue #755, opened 2017-08-22
[Request] [Equalizer] Add option to choose audio device to display on keyboard


diogotr7:
If we could change the audio device it would be possible for the EQ layer to only display music, for example.

It would allow voicemeeter banana users to isolate certain applications while stil playing all the sound to the user.

iimma:
good idea, for example my keyboard light works only with the audio that my monitor is taking out of my hdmi port. i'm not able to use realtek's devices or any other audio device.

diogotr7:
If anyone is going through old issues, you can achieve this with the newest windows update and Virtual audio cable. Force the default output of aurora to virtual cable, then in windows recording devices, select "Listen to this device". If you want to isolate a single application, like spotify, force the output to the same virtual cable.

Issue #1178, opened 2018-07-18
Audio layer preferences


iLeoLion:
### What are you experiencing an issue with?

- [ ] Aurora
- [ ] Aurora supported game, (Fill in the game)
- [ ] Aurora Wrapper
- [x] Aurora Profile
- [ ] Aurora supported device, (Fill in problematic devices)
- [ ] Aurora Scripts
- [ ] Aurora Updater
- [ ] Aurora Lighting issue
- [ ] Other, (Fill in another issue here)

### What kind of an issue is it?

- [ ] Crash
- [ ] Performance problems
- [ ] Visual bug
- [ ] Localization issue
- [ ] Enhancement suggestion
- [ ] Device issue
- [x] Other, (Fill in another type of issue here)

Hi all! I was wondering if in the "Audio Visualizer Layer" was possible to add the option to make it work just for one application (ex. Spotify) and NOT all the audio it captures (ex. Youtube video).
Is it possible?

Thanks!

diogotr7:
I have this exactly setup at the moment. Check out issue #775 and join the discord if you need help!

iLeoLion:
I'm sorry i don't have a problem with the keyboard itself (I can't understand how you know my setup (it's wrong too)). 
I have a problem with the Audio Visualizer Layer :/

diogotr7:
I meant to say #755 , I'm sorry :)

iLeoLion:
Oh nice thanks for the help!
I'll check it out asap

PR #1816, opened 2019-11-25
Audio Device Picker

### LINKED ISSUE #1784: Add ability to change audio output for Visualizer
For people that are actually outputting their audio through a different device than the windows default. For example with Voicemeeter the windows output is just voicemeeters input and the actual output is voicemeeters output

Wibble199:
Adds options to the Aurora general settings for selecting which audio devices are used to provide the values for the mic and speaker volume levels for the `LocalPCInfo` gamestate.

Also adds an option to the audio visualizer to choose which audio output device to use. Different visualizer layers can use different devices.

Adds a utility class for easier handling of switching audio devices. This class can be given an audio device's name and will gracefully switch over to using that device instead. The class keeps track of any events added to the wave's DataAvailable event and will switch these over automatically also.

Resolves #1784.

### DIFF
--- a/Project-Aurora/Project-Aurora/Profiles/GameState.cs
+++ b/Project-Aurora/Project-Aurora/Profiles/GameState.cs
@@ -9,6 +9,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
 using NAudio.CoreAudioApi;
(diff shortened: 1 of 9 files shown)


xQwexx:
I think this closes #755, #1178 Issues as well

PR #1834, opened 2019-12-05
Add stale GH action

Wibble199:
Adds a GH action to mark issues and PRs with the stale label after 90 days (~3 months) of inactivity.
If the issue/PR still has no activity within 2 weeks, it is closed.

Note that the operations-per-run has been set above the default since, at the time of writing, there are a lot of stale issues so we'd like the bot to not spend weeks sorting these. Once the initial ones have been sorted, this could be reset to the default (30) or removed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,23 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


Wibble199:
Temporarily closed until multiple exempt labels can be added (actions/stale/#19).

Issue #68, opened 2016-07-14
GPU Temperature/Usage Displayed on the keyboard


M05T3RT:
Hi antonpup, great work. Got my Logitech G910 this week and was very disappointed until I got Aurora to use my backlighting to it's full potential. I would like to suggest adding GPU indicators on the keys, similar to the CPU and RAM usage. It is not a necessity, but would be really nice to have. Temperature would probably be the most used variable to monitor on a GPU.

Logitech should definitely learn a few things from you in term of backlighting control.

antonpup:
Will see what I can do.

- CPU Temperature
- GPU Temperature
- Mobo Temperature (not sure if useful at all?)
- GPU Usage

M05T3RT:
Awesome, those would all be great. I'm sure someone out there might use motherboard temperature.
Thanks

Pandry:
Maybe the MoBo would not be so useful...

brainbug89:
I use OpenHardwareMonitor to monitor my temps (CPU, GPU, MoBo). Maybe you can use WMI to get the values. So you don't need to add support for every MoBo etc. yourself. :wink:

VoronFX:
Checkout my scripts. There are some features you are asking for, and even more.
https://github.com/VoronFX/Aurora

VoronFX:
Also I am currently working on providing access to all this info from Aurora itself.
In my opinion temp, memory info and a lot of other things are almost useless cause they are auto controlled. What is useful is cpu usage, gpu usage, network connection. 
But anyway I will add as much as I can.

brainbug89:
> Checkout my scripts. There are some features you are asking for, and even more.

Looks nice! :+1: 

I would like to see OpenHardwareMonitor (with [WMI](http://openhardwaremonitor.org/wordpress/wp-content/uploads/2011/04/OpenHardwareMonitor-WMI.pdf)) support for the Percent Effect Layer. So we can display all values (like temps, memory, fan speed, used HDD space etc.) which OHW providing, on the keyboard.

VoronFX:
Work in progress. Need some time.

DogeAmazed:
If we want to get support without using OpenHardwareMonitor w/ WMI, we need to create an wrapper for NVAPI and also for other GPU vendors.
Here's an wrapper for NVAPI: https://github.com/falahati/NvAPIWrapper

VoronFX:
[...]

Issue #1609, opened 2019-06-20
CPU utilization shader will not fully light up at 99% CPU


stdedos:
With layer set as (but `Maximum Value` at 100):
![image](https://user-images.githubusercontent.com/133706/59871803-0b4b1b00-93a1-11e9-9193-43672f25fbc9.png)

Then, even 99.98% CPU will not light up F12

### Expected behavior:

If gradient is done by # of keys, then `CPU >$((MaxCPU - MaxCPU/keys))%` would light up `F12` (i.e. `92.307%` in my `Max CPU 100%` case)
If gradient is done by "actual keyboard distance", then `CPU >$((MaxCPU - MaxCPU/dst_unit))%` (idk how/if you "somehow" measure that)

### Actual behavior:

Only `CPU 100%` would trigger that (and could be not the case, I tried 99,5% and barely got it).

### How often does this problem occur?

This problem occurs... (Fill in here)
Constantly

### Please describe the steps to reproduce this issue:
The issue can be reproduced by doing following steps:

1. 
2. 
3. 

### What is the version of Aurora you are experiencing this issue with?

My version is... `0.6.4b-b604`

### Does the previous version of Aurora have this issue?
~~The previous version, (Fill in the version here), did/did not have this issue.~~ Haven't tried / I guess yes

Mikowmer:
Try removing ESC from the affected keys.

Wibble199:
I don't think the problem is with the logic, as it works fine for other variables, so perhaps the program is reading the CPU usage incorrectly for some reason.

Try typing a value directly into the "Current Value" box, such as '100' and validating whether all the keys light up fully and correctly. I would be surprised if this doesn't happen. If it works as expected when testing raw values, the problem will be with reading the CPU usage values for some reason, but we'll try figure it out once you've tried this.

stdedos:
> Try removing ESC from the affected keys.
> 
> EDIT: Or try to set it up like this (i.e. the default):
> [imgur.com/irtEqo2](https://imgur.com/irtEqo2)

Doesn't work. `CPU @ 99.98%` and still `F12` is unlit

----

> I don't think the problem is with the logic, as it works fine for other variables, so perhaps the program is reading the CPU usage incorrectly for some reason.

My guess would be so. However, it is hard to test it e.g. with RAM utilization (because near-or-100% RAM is probably BSOD).

[...]

Issue #469, opened 2017-02-26
Temperature Display


jdubnz:
### What are you experiencing an issue with?

- [ ] Aurora
- [ ] Aurora supported game, (Fill in the game)
- [ ] Aurora Wrapper
- [ ] Aurora Profile
- [ ] Aurora supported device, (Fill in problematic devices)
- [ ] Aurora Scripts
- [ ] Aurora Updater
- [ ] Aurora Lighting issue
- [x] Other, (Enhancement)

### What kind of an issue is it?

- [ ] Crash
- [ ] Performance problems
- [ ] Visual bug
- [ ] Localization issue
- [x] Enhancement suggestion
- [ ] Device issue
- [ ] Other, (Fill in another type of issue here)

Is it possible to get the keyboard to display CPU and GPU temperature like it does usage? or Fan speed?

Ghanigan:
I feel like this could be possible through Open Hardware Monitor? (http://openhardwaremonitor.org/)

brainbug89:
See #303 and #68 maybe ask @VoronFX he has some nice scripts already: https://github.com/antonpup/Aurora/issues/149#issuecomment-269305453

VoronFX:
Definitely possible, but not yet build in. It will take some time to implement and right now I am too lazy and have other things to do. So I am not contributing at the moment. Maybe someone else will find time and will for this. Or I will contribute again after some break (I hope I will).
IMHO: I consider temp and fan speed as pretty useless values in compare to load. They are auto managed by system and they does no impact except extremely high values. On the other side they may display whether you ready to fry eggs with your PC or no.

brainbug89:
> They are auto managed by system and they does no impact except extremely high values. On the other side they may display whether you ready to fry eggs with your PC or no.

Not at mine. I like to observe the temps to know wheter I need to adjust my water cooling system. :wink:

antonpup:
> On the other side they may display whether you ready to fry eggs with your PC or no.

I absolutely love cooking up eggs or heating up my coffee on my FX8320.

But anyway, I am not opposing the idea of adding this. Adding this won't be difficult anyway.

jdubnz:
Had a play with some of the scripts @VoronFX wrote. Great work. I want to use this on my G810, but obviously I'm missing the Macro G keys.

How should I change the script to get the CPU cores to be represented by the: NumLock   /   *   - keys? 
Cheers!

VoronFX:
@jdubnz , change these lines
[...]

Issue #149, opened 2016-09-20
Per-core CPU usage lighting effect.


YielShelion:
I'm suggesting a new feature for per individual core CPU usage lighting effect so that every CPU has it's own usage meter.
That would be very neat and fun to watch!

antonpup:
I will look into this, but no guarantees.

VoronFX:
Checkout my scripts, there is such feature.
https://github.com/VoronFX/Aurora

Pandry:
@VoronFX Really cool mate!

YielShelion:
@VoronFX it looks cool and great! 
However since I am using razer, I want it to be a progressive bar on keys 1~9 for core one, Q to P for core two, and so on. How do I manage that?

(Sorry for putting this here since I can't ask you directly)

VoronFX:
@YielShelion I will write a new script for you that allows to do what you asking about. Will end up in less than a day.

VoronFX:
# @YielShelion I've just ended an update that includes new script that allow do exactly what you want. Check it out!

YielShelion:
@VoronFX Can I have your contact please? I have something to discuss.

VoronFX:
@YielShelion Sure. 
my email is voron.exe@gmail.com, 
skype is Voron.exe
Also you can open discussions on my repository.
I will delete this info through a while to not get into spam subscriptions.

nyan257:
@VoronFX how do i delete the rainbow effects from your scripts? whatdoes the ping pulser show?
Im sorry for the noob questions...its just that im new to computers.

VoronFX:
@nyan257
To disable rainbow you need just to comment or delete one line:
`layers.Enqueue(CPULayerRainbowCircle);` for CpuCores script and 
`layers.Enqueue(GPURainbowLayer);` for GpuLoad script.

PingPulser shows current ping to desired host by bar, also animation itself represents each ping request.
I hope you know what ping means, do you? By default it pings google.com but it can also ping different hosts in depend on active application. To show ping to game server you currently playing for example. But that needs some additional settings. If bar is full than ping is 400 ms or higher. If there is no connection full bar will blink red twice and fade out.

nyan257:
@VoronFX one other thing... For the cpu cores plugin whenever all the core icons are red all the effects are disabled.  How do i prevent this?  Also how to i change what buttons the ping pulser shows on?

VoronFX:
@nyan257 
[...]

PR #1858, opened 2019-12-28
Added more sensors like CPU and GPU temperatures and usage

diogotr7:

- Added HardwareMonitor in Utils
- This Utility can start and stop updating information on a per hardware basis based on if it is being accessed or not
- Added CPU, GPU, and RAM nodes to LocalPCInfo with useful variables

**Known issues/To do:**

- Some of the variables will be 0 opn some systems because not every hardware has every sensor
- Had to leave the old CPUUsage and Memory variables to not break profiules. Maybe could remove them and just change the path when loading old profiles? 
- Can cause some cpu usage on nvidia optimus systems when reading gpu data. This issue doesnt happen if the gpu is being used, such as when playing a game. I don't know if there is a way to detect this.
- Maybe some place in the UI to choose the update rate of each hardware? Would appreciate ideas.


xQwexx:
I think this is going to fix #68 , #149 , #469 ,
and if it was the incorrect reading then #1609

PR #1892, opened 2020-01-23
Ducky Keyboard Support

### LINKED ISSUE #1391: Ducky Shine 7 support?
Newly released Ducky Shine 7 has Razer Chroma compatibility. The keyboard is detected in the razer synapse program. Ducky has also released a new version of their own RGB software too. Any chance of taking a look at it?

thingamajig0:

Fixes #1391 

- Adds these Ducky keyboards:
  - Ducky Shine 7
  - Ducky One 2 RGB
  - Ducky One 2 RGB TKL

**Known issues/To do:**

- This does NOT include support for the Ducky One 2 Mini. That keyboard can never be supported due to there being no rgb communication between the keyboard and the PC.


Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
```diff
@@ -0,0 +1,352 @@
...
+             and one terminate packet
+             
+             These packets are 64 bytes each (technically 65 but the first byte is just padding, which is why there's the .Take(65) there)
+             Each key has its own three bytes for r,g,b somewhere in the 8 colour packets. These positions are defined in the packetOffsetMap
+             The colour packets also have a header. (You might be able to send these packets out of order, and the headers will tell the keyboard where it should be, but IDK)*/
+            for (int i = 0; i < 10; i++)
+            {
+                try
+                {
+                    packetStream.Write(colourMessage.Skip(Packet(i)).Take(65).ToArray());
```
Wibble199: I'd be tempted to use the overload for Write() that takes an offset and count instead of using LINQ.

thingamajig0: Done :)

thingamajig0:
I've added ISO support now. Just need someone to test. It's difficult finding someone with this keyboard that has the ISO layout, since almost everyone has the ANSI one...

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyRGBAPI.cs
```diff
@@ -0,0 +1,618 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Aurora.Devices.Ducky
+{
+    class DuckyRGBAPI
```
diogotr7: This whole class can be static.

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
```diff
@@ -0,0 +1,360 @@
...
+            colourMessage[Packet(1) + 17] = 0xFF;
+            colourMessage[Packet(1) + 18] = 0xFF;
+            colourMessage[Packet(1) + 19] = 0xFF;
+            colourMessage[Packet(1) + 20] = 0xFF;
+            //Sets terminate colour packet
+            colourMessage[Packet(9) + 1] = 0x51;
+            colourMessage[Packet(9) + 2] = 0x28;
+            colourMessage[Packet(9) + 5] = 0xFF;
+
+            shine7Keyboard = DeviceList.Local.GetHidDevices(0x04D9, 0x0348).SingleOrDefault(HidDevice => HidDevice.GetMaxInputReportLength() == 65);
```
diogotr7: These magic numbers should probably be `const`.

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
```diff
@@ -0,0 +1,360 @@
...
+            //Sets the initialize colour change packet
+            startingPacket.CopyTo(colourMessage, Packet(0) + 1);
+            //Headers for each colour packet
+            for (byte i = 0; i < 8; i++)
+            {
+                colourHeader[2] = i;
+                colourHeader.CopyTo(colourMessage, Packet(i + 1) + 1);
+            }
+            //First colour packet has extra data
+            colourMessage[Packet(1) + 5] = 0x01;
```
diogotr7: i'd maybe throw these magic bytes in a static array called `firstColorPacketData` or something and then Array.Copy it to `colorMessage`.
are the ones you don't set supposed to be zero?

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
```diff
@@ -0,0 +1,360 @@
...
+        public bool UpdateDevice(Dictionary<DeviceKeys, Color> keyColors, DoWorkEventArgs e, bool forced = false)
+        {
+            foreach (KeyValuePair<DeviceKeys, Color> kc in keyColors)
+            {
+                //This keyboard doesn't take alpha (transparency) values, so we do this:
+                processedColor = ColorUtils.CorrectWithAlpha(kc.Value);
+
+                //This if statement grabs the packet offset from the key that Aurora wants to set, using packetOffsetMap.
+                //It also checks whether the key exists in the Dictionary, and if not, doesn't try and set the key colour.
+                if(!packetOffsetMap.TryGetValue(kc.Key, out currentOffset)){
```
diogotr7: I don't think it's clear that `currentOffset` just contains which packet the key is supposed to be sent in, and the index where the colors should go. Maybe storing this as a named tuple instead of a random byte array would be better? would remove the need for the `PACKET_NUM` and `OFFSET_NUM` constants as well.

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
```diff
@@ -0,0 +1,360 @@
...
+            /*
+             The keyboard is only set up to change all key colours at once, using 10 USB HID packets. They consist of:
+             One initializing packet
+             Eight colour packets (although the eighth one isn't used at all)
+             and one terminate packet
+             
+             These packets are 64 bytes each (technically 65 but the first byte is just padding, which is why there's the .Take(65) there)
+             Each key has its own three bytes for r,g,b somewhere in the 8 colour packets. These positions are defined in the packetOffsetMap
+             The colour packets also have a header. (You might be able to send these packets out of order, and the headers will tell the keyboard where it should be, but IDK)*/
+            for (int i = 0; i < 10; i++)
```
diogotr7: You could store the last array you sent and only send it again if the colors changed, maybe that helps with the keyboard not responding after a while.

Project-Aurora/Project-Aurora/Devices/Ducky/DuckyRGBAPI.cs
```diff
@@ -0,0 +1,618 @@
...
+    {
+        private static Dictionary<string, byte[][]> controlCommands = new Dictionary<string, byte[][]>
+        {
+            {"Shine_7_Takeover", new byte[][]
```
diogotr7: Storing this as a static object instead of a value in a dictionary makes more sense to me. that way you'd be able to do 
```cs
foreach(var packet in DuckyRGBAPI.Takeover)
{
//do whatever
}
```
instead of having random strings everywhere.
IMO the class name doesnt make sense. you either store all ducke methods here and call them from the device integration, or rename this something like `DuckyConstants`.

PR #1903, opened 2020-02-07
Replaced Wrapper method with the new ingame api for league of legends

diogotr7:
Riot recently added an api to get in game data so i'm using that instead of the old lightfx wrapper for league (that doesnt do that much anyway).

Things to fix / improve:
* Some way of automatically removing the old wrapper patch from the folder as it makes the web requests not stop sometimes because the pipe remains connected.
* improvements to the default champion colors.
* maybe improvements to the default profile colors?
* some documentation in the layer UI?


diogotr7:
Might be a good idea to wait some time to see if the API changes. It is not documented yet as it is very recent so things could change.

diogotr7:
https://developer.riotgames.com/docs/lol#game-client-api

stdedos:
I am not wishing for "professional" upgrade/downgrade handling, but can there be _something_?

I am getting this (https://github.com/antonpup/Aurora/issues/1609#issuecomment-598041230)
image

and then also that:
image

diogotr7:
The first popup is normal (though i'm noticing now that it says overwatch, which is obviously a mistake) 
After this, i'm not sure what causes it to continue erroring out. This should happen *once* when you update, and no more. Does this happen more than once on your system?

stdedos:
> The first popup is normal (though i'm noticing now that it says overwatch, which is obviously a mistake)

I didn't take a screenshot, as I was not planning to report this. I recycled the one I used for Overwatch.

At least, this time, both popups were only shown once! 😃

Good thinking me, that I didn't open a new issue 👍

PR #1941, opened 2020-03-25
Added spotify support with a spicetify plugin

diogotr7:
Probably a good idea to re-think the installation process.
thanks @th3an7 for the idea and plugin implementation


Project-Aurora/Project-Aurora/Profiles/Spotify/Control_Spotify.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+                return;
+            }
+
+            try
+            {
+                using (FileStream pluginStream = File.Create(pluginFile))
+                {
+                    pluginStream.Write(Properties.Resources.AuroraSpotify, 0, Properties.Resources.AuroraSpotify.Length);
+                }
+                MessageBox.Show("Plugin installed successfully");
```
Bluscream, line 84: Looks like it also has to be enabled with either
- running `spicetify config extensions AuroraSpotify`

or by
- modifying `%USERPROFILE%\.spicetify\config.ini`

and running `spicetify apply`

Bluscream (changes requested):

diogotr7:
This can be done with the official spotify API, so i will not be merging this. I will work on the official spotify api implementation later.

PR #1956, opened 2020-04-11
Update MinecraftApplication.cs

IhToN:
Minecraft title, when it's a modded client, it comes with a "*" on its name.

With this little modification on the regex file the profile will work again on Forge/Fabric modded clients.


diogotr7:
This has been fixed with #1924 . Thanks for contributing, though!

PR #1981, opened 2020-04-26
small minecraft fixes

diogotr7:
adds fabric download link
small tweaks to the default profile


diogotr7:
I was going to replace the background layer as well but honestly i think the effect is specific enough to the game that I'm going to leave it in. It's a bit messy to reproduce the full transition effect with overrides.

PR #1991, opened 2020-05-03
Allow Node funcs to be accessed in outside plugins.

wolfcomp:
Does what the PR title says.
Just a simple thing that should be allowed this will help making plugins in specific ways easier.


diogotr7:
I can't see why this would cause issues so sure :p

PR #1999, opened 2020-05-07
Fix issue that Aurora cannot be launched when the Ducky keyboard is not connected.

SimonFu-HP:
Fix issue that Aurora cannot be launched when the Ducky keyboard is not connected.

- Check the result of opening Duck keyboard


diogotr7:
Duplicate of #1996

PR #2005, opened 2020-05-13
Replaced League of Legends icon with a new one.

PixelHir:

- none

- Replaces the old League of Legends icon with a new one. The old one is no longer used anywhere, while the new one has been extracted directly from the League of Legends executable.
image

**Known issues/To do:**

- none


diogotr7:
I don't know how I missed the new icon... Thanks!

PR #2011, opened 2020-05-14
Uniwill replace try catch

xQwexx:


Tyberious:
Tested on supported system and no issue found.

PR #2024, opened 2020-05-20
Fix invalid enumeration exception when desktop disabled

xQwexx:
Fix use case when deleting the device in the foreach loop.

### DIFF
--- a/Project-Aurora/Project-Aurora/Devices/Asus/AsusDevice.cs
+++ b/Project-Aurora/Project-Aurora/Devices/Asus/AsusDevice.cs
@@ -59,8 +59,11 @@ namespace Aurora.Devices.Asus
         /// <inheritdoc />
         public void Shutdown()
         {
-            asusHandler.Stop();
-            isActive = false;
+            if (isActive)
+            {
+                asusHandler.Stop();
+                isActive = false;
+            }
         }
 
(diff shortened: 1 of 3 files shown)


Fluto:
`asusHandler.DisconnectDevice(this);` should still be called

Every piece in the code that uses a foreach to iterate over the devices should use the lock, and should be a for loop instead. In the case where something may be removed, the for loop should be reversed.

For example 
`Project-Aurora\Devices\Asus\AsusHandler.cs : 181`
```CSharp
foreach (var device in devices)
                    device.Stop();
```
Should be this instead
```CSharp
for (var i = devices.Count - 1; i >= 0; i--)
                    devices[i].Stop();
```

xQwexx:
At first I also thought that the asusHandler.DisconnectDevice(this)
Should be called in the taskcanceledexception but after think it through, this exemption is going to happen if the running thread was cancelled via device.Stop() and every time when this was used it's followed by device deletion and that's all what is the disconnectdevice function is doing (apart from logging)

diogotr7:
It would be awesome if we could get this fixed and merged as it looks like a significant issue.

diogotr7:
#2037 has a couple more fixes so i'll merge that one instead. Thanks for contributing

Issue #1975, opened 2020-04-23
Multiple Audio Input Devices support


Bluscream:
Some people don't have a "Stereomix" output device on their PC and therefor always need to switch between devices. It would be better if there was a listbox where we could check all the devices we want aurora to listen/respond to

PR #2028, opened 2020-05-21
Audio device picker

Wibble199:
Adds the ability to select which audio device to use for individual audio visualizer layers and globally for the local pc information game state.


Bluscream:
Fixes #1784 and #1975

diogotr7 (approved):

Issue #1953, opened 2020-04-07
GHUB keeps overwriting Aurora


trxraptor:
### Expected Behaviour
Keyboard lighting switches between Aurora profiles when changing apps
### Actual Behaviour
(Logitech G910)
At Windows startup when Aurora starts (with some delay to ensure GHUB is open), lighting is GHUB profile. To make Aurora work I need to close Aurora and open Aurora again. Then Aurora works until I switch to a program with a profile (a game for example). When I switch to the game, the lighting is GHUB. When I go back to the desktop, the lighting is still GHUB. I can fix it temporally by going to Settings->Device Manager then stopping and starting Logitech. However, right after Windows startup, under Settings->Device Manager, Logitech is seen as disconnected and the stop start method doesn't work until I restart Aurora.

### Frequency
Every time

### Aurora Version:
beta v0.7.2 (GHUB Version: 2020.2.35921)

### Previously an Issue?
No (Before the latest GHUB update)

samb10132:
Same thing is happening to me. If you set your Logitech G HUB to a persistent profile (I did default desktop) I believe that fixes the problem for the most part.

TrueShack:
also, in my case, while I got the GHub lighting with Aurora running, GHub still says the lighting is controlled by Aurora and Aurora shows the lighting that should be on the keyboard in realtime.

Norbo11:
I have the same issue. My PC wakes up from sleep and my keyboard is showing the GHub profile (I have selected the default profile, a solid light blue across the keyboard). I have to close Aurora, which doesn't close it properly, so I have to kill the Aurora.exe process from the Task Manager. Starting Aurora again puts it in the Aurora profile.

> Same thing is happening to me. If you set your Logitech G HUB to a persistent profile (I did default desktop) I believe that fixes the problem for the most part.

I have my GHUB set to a persistent default profile and it doesn't fix anything.

trxraptor:
> Same thing is happening to me. If you set your Logitech G HUB to a persistent profile (I did default desktop) I believe that fixes the problem for the most part.

I use GHUB macros heavily, so a persistent profile won't work for me.
I also tried the Override DLL option, but it didn't made a difference.
In Patch Notes for GHUB update 2020.02, it says

> Fixed an issue with the LED SDK

So the problem is probably caused by the new GHUB update.

trxraptor:
[...]

PR #2054, opened 2020-05-30
Added return of actual operation in LogiLEDWrapper

supergrml:
Related to: #1953

- The wrapper returns the actual operation status combined with the init status. This gives the application the chance to check if the device and/or Logitech software is still present.

**Known issues/To do:**


diogotr7:
I dont understand what you're trying to solve here? The wrapper is for logitech **games**, not devices. This is never used for connecting to LGHUB or LGS from Aurora. The purpose of the wrapper is acting as a replacement to LGS to the eyes of, for example, GTA 5. This allows aurora to capture the lighting intended for logitech devices and display it somewhere else.

supergrml:
If you lose connection to the device or GHub/LGS the wrapper doesn't return an appropriate result. The LogitechLed gives back the right status, this change is meant to be consistent.

diogotr7:
This issue does not solve #1953 at all but the wrapper changes do make sense...

I'll take the opportunity to fix an issue i found with the wrapper light event while testing the changes.

Thank you for contributing.

PR #2065, opened 2020-06-03
OEM list renew.

rushdie99:
OEM list renew.


gitmacer:
@rushdie99 Can we get a OEM/device list? So we can add it to the wiki. https://github.com/antonpup/Aurora/wiki/Supported-Games-&-Devices#supported-devices 

Being able to check if your device is supported before you download aurora/buy a device would improve the user experience a lot. Also there are many people that are actively searching for aurora compatible devices so I am pretty sure it would even improve sales. Especially because near no one of the users know the company Uniwill a list is needed.

PR #2072, opened 2020-06-06
Add boolean change and T flip-flop evaluatable

Wibble199:
This PR adds two new evaluatables:
- A boolean change detector that works similarly to the numeric change detector. Will output a true signal on the frame that the child evaluatable value changes. Can be configured to only output when the new value is true or when the new value is false.
- A toggle flip-flop whose state changes every frame that the input evaluatable is true.

The existing "_Flip-Flop_" (`Boolean_Latch`) has been renamed to "_Flip-Flop (Set-Reset)_" (`Boolean_FlipFlopSR`) to clarify the type of latch it behaves as.


Project-Aurora/Project-Aurora/Settings/Overrides/Logic/Boolean/Boolean_FlipFlop.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        private bool state = false;
+
+        public Evaluatable<bool> Toggle { get; set; }
+
+        public Boolean_FlipFlopT() : this(EvaluatableDefaults.Get<bool>()) { }
+        public Boolean_FlipFlopT(Evaluatable<bool> toggle) => Toggle = toggle;
+
+        protected override bool Execute(IGameState gameState) {
+            if (Toggle.Evaluate(gameState))
```
gitmacer, line 25: It should only toggle if it got false since last toggle. I know you can "fix" it with the boolean change detector but I still think this change should be done. Everyone that worked with graphical programming before will expect it works like that.

gitmacer, line 25: It should only toggle if it got false since last toggle. I know you can "fix" it with the boolean change detector but I still think this change should be done. Everyone that worked with graphical programming before will expect it works like that.

diogotr7: I think the current approach makes the most sense and is the most flexible. I don't consider adding a change detector to be a 'fix', for both options to be available, you'd need to add something one way or the other. This way it's more clear what each evaluatable does.

PR #2087, opened 2020-06-11
Fixed issue where desktop overlays are rendered multiple times.

Wibble199:
Fixed an issue where overlay layers on the desktop profile were being updated twice in a single frame when either the Aurora window was not focused or the desktop was being previewed while "Show Overlays in Application Preview" is checked.

This was particularly noticable with the particle layer. Creating one as a normal layer and one as an overlay demonstrates the issue.


gitmacer:
@Wibble199 This PR-Broke Overlayer UI preview.

PR #2097, opened 2020-06-14
Fix effect layer constructors being slow because setting new Bitmap is slower than Clone.

wolfcomp:
Speeds up `new EffectLayer()` and those constructors

Without this code prof

With this code profoptim

Both of these are taken over an estimate of 20 seconds note the difference in self CPU usage for the `Aurora.EffectsEngine.EfectsLayer::.ctor` _second percentage column_

All of this is using default settings for the bitmap accuracy.


diogotr7:
please dont forget these changes, i wont merge then right now though. When we make the merge to the new repo i want to do the optimizations talked about on the discord. I'm afraid there's some weord interactions we aren't seeing that break profiles. 

When we move repos properly i'll ask you to submit again :)

PR #2107, opened 2020-06-20
added checks of retrun codes from logitech sdk

supergrml:

Fixes: #1953

- added an option to deinit and release the LGS/GHub dll
- check if the sdk functions actually succeed
- Aurora starting after LGS/Ghub
- Connection to LGS/GHub lost (program closed/restarted)
- detaching/reattaching a device

**Known issues/To do:**

- when switching profiles in GHub, Ghub sometimes rejects sdk commands

### DIFF
--- a/Project-Aurora/Project-Aurora/Devices/Logitech/LogitechDevice.cs
+++ b/Project-Aurora/Project-Aurora/Devices/Logitech/LogitechDevice.cs
@@ -152,6 +152,7 @@ namespace Aurora.Devices.Logitech
     {
         private String devicename = "Logitech";
         private bool isInitialized = false;
+        private bool ledSDKPresent = false;
 
         private bool keyboard_updated = false;
         private bool peripheral_updated = false;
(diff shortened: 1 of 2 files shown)


Cheerpipe:
Tested with several logitech devies (kb, mice, speakers and mousemat).

It works (randomly) for the keyboard but not for devices that takes more time to be initialized by G HUB.

Test case:

1. Open G HUB
2.  Open Aurora

Every is working fine so far

3. Close G HUB

Logitech device stops in Aurora Device Manager

4. Start G HUB Again

Keyboard is synced with Aurora but mouse and mousemat (speakers some time) are out of sync. Maybe because Logitch SDK is re-loaded before G HUB detects all devices.

diogotr7:
I rewrote the logitech integration recently (dev branch), it should have this feature in it. If you notice that it does not work correctly, please submit a new pull request. Aurora seems to detect when LGS is closed (since the Set methods return false when that happens). I think that's what you were trying to solve.

Thanks for bringing this issue to my attention.

PR #2109, opened 2020-06-20
Convert .csproj files to net SDK format

ChrisDurante:
- Convert managed projects to use NET.Sdk project file format while maintaining the current build process
- Remove dependency on `SharpZipLib` in favor of standard .NET implementation in `System.IO.Compression`
- Remove `AssemblyInfo.cs` AssemblyFileVersion replacement in appveyor.yml build with MSBuild tooling.
  - The version passed in the appveyor pipeline takes precedence in the `Directory.Build.props` file for Project-Aurora

The changes proposed should make it easier to maintain the WPF projects, since file adds and deletes won't modify the .csproj file. It may also help with any future development to move from full .NET Framework to .NET Core.


diogotr7:
When i was upgrading the project to .net framework v4.8 i tried to do this conversion but it caused more problems than it was worth.

I've noticed a few issues with this: the version from appveyor builds should contain the the version *and* the build number. For this build, the version should be 0.8.0+875 and it appears to be just 875.

Thanks a lot for working on this, it really was needed.

ChrisDurante:
Glad to help! I tried a few different things in the build.props file, but it seems like MSBuild doesn't find the appveyor version number, which is strange because previously the powershell task to replace the file version used the environment variable. Hopefully I can figure out why. Alternatively, there are different ways to patch the versions in the appveyor docs but I didn't intend to change the build that much.

ChrisDurante:
Hi @diogotr7 - I was able to match the versioning scheme by removing the generation of the informational version attribute. This should be ready for review. In the future versioning and build can be revisited, but it's not necessary to change as part of this PR.

diogotr7:
I merged a bunch of things today and this conflict was getting a bit ugly so i decided to handle it sooner rather than later. I tried to group things the same way you had them already.

I hope i didn't ruin anything, let's see if appveyor likes the build :)

diogotr7:
I'd appreciate it if you would take another look, i might've missed something in my changes. Otherwise, i want to merge this.
The appveyor build id seems to work fine still.

image

diogotr7 (dismissed):

ChrisDurante:
Hi @diogotr7 - nothing missed in the merge, I know they get a lot harder if contributors add features or restructure the project. 
I was thinking of moving all the included .dll's into a folder other than the root of the project but figured that could be done later. I know the ordering is a little weird for those non-referenced .dll's - alphabetical with root folder first. I also made the spacing consistent (could switch to tabs for all csproj files, but I think the default is spaces and we don't have an .editorconfig)

diogotr7:
Thanks for contributing! Let's hope this doesn't bring any unwanted behaviour i'm not seeing yet...

diogotr7 (approved):

Issue #1160, opened 2018-07-02
PD2 Preview tab always active


Practiclycrp:
May be related to #844

### What are you experiencing an issue with?

- [ ] Aurora
- [X] Aurora supported game, (Payday 2)
- [ ] Aurora Wrapper
- [ ] Aurora Profile
- [ ] Aurora supported device, (Fill in problematic devices)
- [ ] Aurora Scripts
- [ ] Aurora Updater
- [ ] Aurora Lighting issue
- [ ] Other, (Fill in another issue here)

### What kind of an issue is it?

- [ ] Crash
- [ ] Performance problems
- [X] Visual bug
- [ ] Localization issue
- [ ] Enhancement suggestion
- [ ] Device issue
- [ ] Other, (Fill in another type of issue here)

### Expected behavior:
Explain here what should happen.

### Actual behavior:
Explain here what happened instead.

### How often does this problem occur?

This problem occurs... (Fill in here)

### Please describe the steps to reproduce this issue:
The issue can be reproduced by doing following steps:

1. Launch Payday 2
2. Start a heist
3. Take damage or use ammo
4. F1-12, 1-=, neither of them update.

It seems like the Preview tabs settings for the game are always in effect while the profile is loaded, causing issues with health & ammo, has the same issue when Game state, Level phase, Player state, are set to something aswell, overwriting whatever may actually be happening ingame, they work fine when set to undefined.

### What is the version of Aurora you are experiencing this issue with?

My version is... (Beta 0.6.1 -dev4)

### Please include any logs, screenshots, or files that are related to this issue:

https://img.prccrp.org/12sqtcY.png
Keyboard is G910 QWERTY

RaDiaTiionS:
i have the same problem. i installed everything it said so dono what to do now

Wibble199:
I've done some digging and I think the reason some of the PAYDAY 2 features aren't working is because the gamestate integration mod for PD2 is out of date and doesn't send all the required information to Aurora.

I know this isn't the best response as it means there's no fix in sight (until someone more knowledgeable than I about PD2 modding can fix the mode).

Wibble199:
For any devs curious, this is the data that is sent from the PD2 GSI mod to Aurora.

![devenv_2018-07-03_10-43-36](https://user-images.githubusercontent.com/3984322/42214195-bb67059e-7eb2-11e8-92a9-3df9cfdcc245.png)

[...]

PR #2134, opened 2020-07-11
Re-Write Payday 2 Effects - Fixes #2133 & #1160

### LINKED ISSUE #2133: Rewrite Payday 2 Integration to be smarter and use more effects
Currently the Payday 2 integration is pretty weak and uses many unnecessary custom layers.

I've set out to re-write the Payday 2 effect layers from the ground up to be smarter, show more effects and react smarter to ingame effects. I've found some limitations with the GSI hook and am aware of strong clashes with other mods [WolfHud], however I've built this from Vanilla.

stuntguy3000:
**List any issues that this PR fixes:** Fixes #2133 & #1160

**This pull request proposes the following changes (with NO client GSI update required):**

- Remove custom layer effects from defaults
- Add new layer effects
- Add new game-responsive effects
- Update instructions and mod download links
- Remove broken preview function

**Known issues/To do:**
N/A


diogotr7:
I'm not sure removing layer sis a good idea. Removing them would break everyones profiles. Changing the default to a regular layer with overrides would make sense though.

stuntguy3000:
Will address your comments/merge issues soon.

stuntguy3000:
> I'm not sure removing layer sis a good idea. Removing them would break everyones profiles. 

Fair point @diogotr7, I've addressed this above (and the merge conflict). As far as I am concerned, this is good to go.

Issue #2027, opened 2020-05-21
WLED (ESP-8266)


ikrowni:
Please add support for the following device:

**Device Name:** WLED (ESP-8266)

**Device Website:** https://github.com/Aircoookie/WLED/wiki

This is heavily integrated with my current office setup through home assistant. I would love the ability to control the lights with my games. 

https://www.youtube.com/watch?v=es9rawSGUPk

Cheerpipe:
I have some Chinese led strips that talks ledenet integrated with Aurora. Did you know what protocol talks your LEDs?

ikrowni:
udp

Blynk

Mqtt

Hue

E1.31

Are all options available within the sync options.

Not sure if any of that helps

ikrowni:
> Did you know what protocol talks your LEDs?

Does this page help?

https://github.com/Aircoookie/WLED/wiki/Add-own-functionality

Operation760:
I second this for a formal enhancement and integration support for WLED. Support for WLED alone opens up some very cool, highly complimentary features/functions and is worth it alone. However, an integrated Aurora-WLED could extend Aurora into a larger home automation ecosystem and open up a lot of very cool possibilities for event based actions and synchronized lighting via applications and platforms like Home Assistant, ESPHome, Tasmota, etc. Possibilities for devices and services far beyond PC peripherals and the Razer-Hue-Aurora integration. 

Think Aurora changing PC lighting behavior to notify you that  someone or something has  triggered motion at your front door or the wife, via presence detection has not tripped a defined geozone exit alert and therefore game time has been extended ;). PC getting too hot? Game room temp increasing? Turn on the AC! 

Tons of possibilities! Also Aurora’s integrations could extend into the WLED and HA ecosystems. Being able to extend PC-basest conditions and events is a sought after use-case and within Aurora’s excellent existing capabilities.

ikrowni:
The developer of the wled project just reach out and said that the JSON API would probably be the best option for an integration into Aurora.

https://github.com/Aircoookie/WLED/wiki/JSON-API

nzbr:
I have created this script a while back for using my RGB led strips with Aurora. It was not created with WLED in mind, so it only uses the UDP interface but it works great for me.

To use it, just put the script in `C:\Program Files\Aurora\Scripts\Devices` as `UDP.cs`. Then, when you restart Aurora, a new device-type UDP should appear, where you can input the IP Adresses.

[...]

PR #2145, opened 2020-07-24
Add support for LED strips using UDP (e.g. WLED)

nzbr:

Support is added for devices that receive color information via colors sent as raw bytes over UDP.  
This protocol (which is called rawudp by [Hyperion](https://github.com/hyperion-project/hyperion.ng)) is supported by WLED, a software for controlling led strips (#2027).

**Known issues/To do:**

The ESP platform (on which WLED runs) only supports up to 490 LEDs using this protocol (see https://github.com/Aircoookie/WLED/issues/1019). To get around this, direct support for WLED using their JSON-API over HTTP would be a solution


ikrowni (dismissed):
Seems to work great on my end and others have had success with it so far. As user nzbr said this would be much better with direct integration through the JSON-API

diogotr7:
Hey, sorry i'm taking so long to get to the PR's.

This needs updating to the new IDevice interface, should be very simple to do: just change the GetSomething() methods to properties instead. Visual Studio's refactoring should do this correctly. Also manually adding the device in DeviceManager is not needed anymore.

As for the http json api, would this be implemented on top of the current code or would it be another Device?

nzbr:
The WLED API would be another device because it would _only_ work with WLED whereas the UDP device is much more generic

I have converted all the getters to properties. It should still work, but I can not test it, as Aurora crashes when I try to set a keyboard model (In `KeyboardLayoutManager.cs:830`)

diogotr7:
weird, I didn't experience any keyboard layout issues. I can't test this properly as i don't have the required hardware but it runs fine otherwise. I don't know why the nzxtsharp submodule keeps coming back, but that's what's causing the CI build to fail. I can remove it if you'd like. Otherwise looks good.

nzbr:
Ok, so I deleted `%appdata%\Aurora` but I am still getting this:
grafik

nzbr:
Ok something must have gone wrong with my repo because I just cloned a fresh version, put `UDPDevice.cs` and it works now

diogotr7:
I noticed this device always shows up as Initialized. Should it be disabled by default? is the initialization code supposed to work like that?

diogotr7:
This still shows up as initialized always. I think disabling it by default makes sense.

SmAsH-jl:
Could we revisit this and instead of UDP consider e1.31 that would give the flexability of ANY rgb pixel device.  strips strings bulbs whatever.

nzbr:
I have now changed the initialization behavior so that the device only initializes if at least one IP is specified. Also, ports are now set by writing `IP:port` instead of setting one port globally, so that it is now possible to have different ports on different hosts.

Having E1.31 support (as an addition) sounds like a good idea to me, though I have no idea how that protocol works tbh.

I had started working on a WLED-API device integration with auto-discovery of WLED devices on the local network some time ago, but I have lost the code for that. I _may_ start working on that again soon, but I can't promise anything

snickers2k:
hey guys. any updates on this? is this actually working? would love to integrate my multiple wled's with aurora.
thanks

nzbr:
The project seems to have been moved to https://github.com/Aurora-RGB/Aurora 

I have not really touched this since my last comment, however it still seems to work on the latest version. If you want to give it a try, i packaged a version of Aurora that includes the UDP device: https://s3.eu-central-1.wasabisys.com/nzbr/Aurora-setup-v2-nzbr.exe
Although not being exactly the same code, the limitiations of the script that i posted in #2027 (only up to 490 LEDs and a single color) still apply here

EDIT: Aurora will try to update to the latest official release, you have to click cancel in the updater to keep the patched version

snickers2k:
thank you ! @nzbr

PR #2172, opened 2020-09-08
Support for multiple YeeLight devices

snake-4:

- Added support for multiple YeeLight devices.
- 
- 

**Known issues/To do:**

- None.
- 
-


diogotr7 (approved):

diogotr7:
1. This breaks old configurations, right?
2. Is the auto discovery not good enough? Why is a manual ip address input needed? 
3. I think just the checkbox being ticked is enough to enable auto discovery. You do not need to check for the ip box to be empty as well.

snake-4:
1. It shouldn't break old configurations as there are no removals or additions to the variable registry. The only thing that has changed is that now it tries to split the value of IP address string using comma as the delimiter.
2. Auto discovery might not always work because some routers might not support multicast and sometimes the light doesn't respond to the discovery requests.
3. The way auto discovery works is that it sets the value of the IP variable in VariableRegistry according to the discovered lights and doesn't do auto discovery if the IP variable is non-empty in order to avoid doing discovery on each start if lights have already been found.

Hukuma1:
+1 to please have ability to input IP address! Seems less error prone, especially if you know the static address on your home network.

Btw Yeelight just released official pairing to Razer Synapse. Have you seen it, @SnakePin? Their solution requires additional app to run, plus it doesn't have ability to set custom IP address. And your desktop has to be connected to same network as the lights, e.g. wifi. Very limiting I would say compared to your plugin. Just thought maybe there was a way to see if they simplified or improved API for you to tap into for your plugin? Particularly interested if they addressed 0 brightness to not close the connection to maintain further control over the bulb.

https://www.yeelight.com/pages/razer.html

yeelight-gallery-01

PR #2195, opened 2020-10-08
Handled default audio device changed

valarion:

- Handle when the default audio device changes, so volume and mute are always linked to the current default audio device instead of the one present when the app starts
- 
- 

**Known issues/To do:**

- None
- 
-


diogotr7 (approved):

diogotr7:
Thank you for contributing!

PR #2208, opened 2020-10-22
Make contributers link read better

boehs:


boehs:
The period bit is following the convention all the other URLs in the about section do not have a period.

PR #2217, opened 2020-10-29
Improve Unified HID device

MatrixDJ96:
Hi Aurora Teams,

I have done some changes in Unified HID devices.

- Fix Asus Pugio bottom led
- Update HID libray
- Add update interval settings in ms
- Add shutdown color (with enable/disable option)
- Optimized CPU usage (set led only when necessary)
- Code cleaned up (and refactor)

**Known issues/To do:**
- Nothing!

Those changes should be merged with pr #2220 to enable shutdown color effect on Aurora closing

### DIFF
--- a/Project-Aurora/Project-Aurora/Devices/UnifiedHID/AsusPugio.cs
+++ b/Project-Aurora/Project-Aurora/Devices/UnifiedHID/AsusPugio.cs
@@ -1,4 +1,4 @@
-﻿using HidLibrary;
+using HidLibrary;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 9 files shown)


snake-4:
Looks okay but I don't see how the update interval setting is useful especially when the default value is 25ms as the Aurora updates at 30FPS anyways.

MatrixDJ96:
> 
> 
> Looks okay but I don't see how the update interval setting is useful especially when the default value is 25ms as the Aurora updates at 30FPS anyways.

I've added the refresh interval option because hidlibray has high CPU usage.
This option allow user to set a slower led update to "fix" this problem

(The 25ms are added to "30FPS" interval)

UPDATE: I've changed default interval to 0ms
UPDATE 2: I've added also shutdown color settings

snake-4:
I feel like shutdown color setting is a bit unnecessary and can be hard-coded to be black instead or at least it can be device specific like the enable setting.

MatrixDJ96:
I could add option to enable/disable shutdown color if for you is "a bit unnecessary" but no hardcoded color, let user decide

UPDATE: I've added option to enable or disable shutdown color

MatrixDJ96:
@diogotr7 @SnakePin this pr is ready for review!

PR #2246, opened 2020-12-31
Fix ducky response lag

### LINKED ISSUE #2086: Ducky Response Lag When Not Debugging
### Expected Behaviour
~30ms device response time

### Actual Behaviour
~154ms device response time

### Reproduction steps
When debugging Aurora, I get around a 30ms device response time and no visual lag. When I run Aurora without debugging, I get around a 144ms device response time with a lot of visual lag. This only affects the keyboard, both my mouse (Logitech G703) and the bitmap preview have no visual lag.

Screenshot (1)
Screenshot (3)

### Frequency
This happens every time I launch Aurora.

### Aurora Version:
I'm using the current code base (Beta v0.8.0).

### Previously an Issue?

This issue is prevalent in all past versions since the Ducky keyboard support has been added.

umbreon222:
**Fixes issues:**
Fixes #2086
- Thread sleep is only guaranteed to wait for _atleast_ the delay time specified but can end up waiting longer ([more on this](https://stackoverflow.com/a/8241018/3775593)). For some reason, thread sleep was waiting a lot longer when not debugging which is causing lag when updating the device.
- Without thread sleep, the device hangs.
  - Added stream flushes to ensure full commands are sent
  - Added input report read after device updates

- Move thread sleep call
- Flush hid stream between commands
- Read input report after device updates

**Known issues/To do:**
- Ducky device hung randomly before adding stream flushes and input report read logic
  - After reconnecting the device, it regained responsiveness
  - Since this occurred randomly I have no way to try and repro it to see if it is truly fixed

### DIFF
--- a/Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
+++ b/Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
@@ -194,7 +194,7 @@ namespace Aurora.Devices.Ducky
                     }
                 }
                 colourMessage.CopyTo(prevColourMessage, 0);
-                return true;
+                Thread.Sleep(10);
             }
             return true;
         }
--- a/Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
+++ b/Project-Aurora/Project-Aurora/Devices/Ducky/DuckyDevice.cs
@@ -185,7 +185,7 @@ namespace Aurora.Devices.Ducky
                 }
                 catch (Exception ex)
                 {
-                    Global.logger.Error(ex);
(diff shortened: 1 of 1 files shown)


umbreon222:
Hold on with this one actually. Lighting seems to freeze and device becomes unresponsive randomly during use.

umbreon222:
This is turning out to be a bigger issue than I hoped for. I can't think of anything else that would remedy the device hanging issue other than a rewriting the support for ducky devices.

Lexevolution:
I'm pretty sure I have fixed the issue in #2263. It also explains how I fixed it. It was some pretty annoying timing issues with the Windows OS.

PR #2248, opened 2021-01-04
OpenRGB More Keyboard Binds

craftycorvid:
- Media Volume and Calc binds for OpenRGB integration
- Proper binding of DPI to PERIPHERAL_DPI
- Don't skip devices that report "Static" instead of "Direct" control in OpenRGB

### DIFF
--- a/Project-Aurora/Project-Aurora/Devices/OpenRGB/OpenRGBDevice.cs
+++ b/Project-Aurora/Project-Aurora/Devices/OpenRGB/OpenRGBDevice.cs
@@ -153,7 +153,7 @@ namespace Aurora.Devices.OpenRGB
             {
                 //should probably store these bools somewhere when initing
                 //might also add this as a property in the library
-                if (!_devices[i].Modes.Any(m => m.Name == "Direct" || m.Name == "Static"))
+                if (!_devices[i].Modes.Any(m => m.Name == "Direct"))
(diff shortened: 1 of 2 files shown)


CalcProgrammer1:
Allowing Static mode is a bad idea!  If a device has a mode that meets all the qualifications to be used with an effects engine, it is given a Direct mode for this purpose.  If a device has only Custom/Static, it means that the device is unsuitable for effect engine use - usually because the mode or color changes save to the device's non-volatile memory which can quickly damage the memory when rapidly updated, as flash memory has a limited write cycle before it starts to corrupt.

If you want to use Aurora with a Static mode device, you're doing so at your own risk of damaging the device.  My RX580 Nitro+ saves all color updates so I named the mode Static on it, but it still seems to work smoothly with effects engines.  I knew the risk and decided to let it run, days at a time.  Now it no longer saves the correct color across power cycles, usually loading up as a white color, almost certainly due to flash wear.  Fortunately for the Nitro+, it can still be sent correct colors after booting, so I don't mind continuing to use effects engines with it, but some devices save and load the colors from the flash resulting in a flicker.  I would guess that these devices won't fail so nicely, since the applied color is read back out of flash.

craftycorvid:
Thanks for the background info. I'll remove that part from the code and leave the rest.

PR #2258, opened 2021-01-14
Logitech SDK update

MatrixDJ96:
- Update Logitech SDK to v9.0
- Removed override dll settings (useless with this SDK)

**Known issues/To do:**
- Nothing!

**Tested device:**
- Logitech G513
- Logitech G502 Hero

**Tested software:**
- G HUB v2020.12.9532 
- Gaming Sofware v9.08.65


diogotr7:
Aurora purposefully has the hacked together split dlls method to get around the led engines wrapper dll. This allows aurora to capture lighting originally meant for Logitech devices like bf1 and gtav and redirect it to something else. The current sdk version is already v9.0

MatrixDJ96:
> 
> 
> Aurora purposefully has the hacked together split dlls method to get around the led engines wrapper dll. This allows aurora to capture lighting originally meant for Logitech devices like bf1 and gtav and redirect it to something else. The current sdk version is already v9.0

Oh sorry I thought that the wrapper was missing because the sdk in use was not the latest version.

MatrixDJ96:
@diogotr7 the dll override is used for what?

diogotr7:
The override shouldn't be needed but it's for when aurora fails to detect whether ghub is running or not. Or to override which dll to use if both are running.

PR #2259, opened 2021-01-16
Add support for loading dependencies of dll devices

sstamoulis:

- Use custom AppDomain.AssemblyResolve event handler to load dependencies of dll devices.
It searches for the dependencies in the same directory the device's dll is located.


diogotr7 (approved):

diogotr7:
Thank you for contributing.

Issue #2188, opened 2020-09-27
Aurora Defaulting Ducky One 2 to profile one


28lachie:
Hi, Auroras great and I love the fact that the ducky one 2 has full support but there is just one thing that does not seem to be working but I could be missing something.

If I have missed the answer I'm sorry but I couldn't find any issues or posts about it.

Basically macro support doesn't work on the ducky because aurora seems to be forcing profile 1 on the keyboard and only profile 2-5 works for macros so I was wondering if macro functionality is still possible? Like it makes sense that it would if you could change to profile 2. Anyway thanks if I'm missing something I'm sorry

Lexevolution:
Unfortunately when any Ducky keyboard is being controlled by the PC, macros are disabled. This was by Ducky's design, however, I have recently made a [Pull Request](https://github.com/antonpup/Aurora/pull/2263) which makes it so whenever you turn Aurora off or stop the lighting, it will go back to the profile you set it to.

PR #2263, opened 2021-01-22
Fix Ducky lag

Lexevolution:

**Fixes:**
- #2086 
- Partial fix of #2188 

- Implements new takeover and release code from the [Ducky Keyboard spec](https://gitlab.com/CalcProgrammer1/OpenRGB/-/wikis/Ducky-Keyboards)
- Fixes timing code (`Thread.Sleep(2)` changed with Hi-Res clock of Stopwatch)
- Update contributors list for new username.

**Known issues/To do:**

- No observable bugs/things to do.
- #2188 is an inherent design flaw of the Ducky keyboards, however, when Aurora exits/releases device from control, it will no longer default to profile 1.
- #2086 was happening because of how the Windows system clock works. `Thread.Sleep` uses the system clock, which has a [default resolution of ~15ms](https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.sleep) ([See also](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer?view=net-5.0#remarks)). This was the cause of the ~150ms per `UpdateDevice` (around 6fps). The Stopwatch uses (by default) a [Hi-Resolution timer](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch#remarks) which updates 10 million times per second. This is why it was used.


Crapy:
Seems to work, should be commited

Lexevolution:
Because the Continuous Integration build is no longer available, you can either build the version with the fix yourself: https://github.com/Lexevolution/Aurora/tree/fix/ducky-lag
Or ask me on the Aurora Discord or DM `Lexevo#0876` to build it and send to you.

diogotr7 (approved):

diogotr7:
Thank you for contributing.

PR #2278, opened 2021-02-21
Proper Logitech G815 support

yener90:


patate1684 (approved):

patate1684 (approved):

Bounty38:
I think He means that keys like pause, next/previous song, mute and keys that change presets are not avaible to configure in Aurora. I Suppose that G915 has same problem

patate1684:
> I think He means that keys like pause, next/previous song, mute and keys that change presets are not avaible to configure in Aurora. I Suppose that G915 has same problem

Yes. Unfortunately this commit doesn't solve the problem. I thinkered with logitech sdk directly and it don't have value for these keys to be change individualy. But there is a workaround. You can change all keys at once with a method of the sdk. This change the color of play/pause, next/previous keys as well. After that, you can change other key and play/pause, next/previous keys stay the same.
So a fix can be done but unfortunately I don't have enough knowledge in programming to push a fix myself.

Bounty38:
> > I think He means that keys like pause, next/previous song, mute and keys that change presets are not avaible to configure in Aurora. I Suppose that G915 has same problem
> 
> Yes. Unfortunately this commit doesn't solve the problem. I thinkered with logiteck sdk directly and it don't have value for these keys to be change individualy. But there is a workaround. You can change all keys at once with a method of the sdk. This change the color of play/pause, next/previous keys as well. After that, you can change other key and play/pause, next/previous keys stay the same.
> So a fix can be done but unfortunately I don't have enough knowledge in programming to push a fix myself.

Well, there is much simpler solution. If you have logitech Ghub running, you can change keys there and it'll remember it's state after launching aurora.

PR #2314, opened 2021-06-02
Upgrade Colore library to 6.0.0

### LINKED ISSUE #2284: razer chroma support disabled
unfortunately razer has updated their razer chroma sdk and now its disabled on aurora.  can someone help with this?

Screenshot (23)

this is on every update from v0.8.1 to v0.8.1 +901

lucasvanhalst:
This should fix the razer chroma support. Tested with razer deathadder V2 and razer firefly

Fixes #2284

Since I'm no professional C# dev I'm not sure what the best way to handle the async stuff is, for now I just used `task.GetAwaiter().GetResult()` which seemed to work fine


diogotr7:
I'm not sure if this will cause deadlock issues because of async. I've fixed the razer support in dev without upgrading to colore v6, so i won't merge this unless another issue comes up. Thanks for contributing though!

Issue #2300, opened 2021-05-12
Gamestate Listener could not start


ghost:
### Expected Behaviour
Aurora to start up

### Actual Behaviour
On load I get an error message - "GameStateListener could not start. Try running this program as administrator"

### Reproduction steps
Fresh install (either zip or exe)
Double click on desktop icon
See error message

### Frequency
Everytime

### Aurora Version:
v0.81

### Previously an Issue?
First time trying

Things I've tried already:
- Running as Administrator
- Running the CLI lines to open up the local host ports (can't remember them off the top but I found them on here)
- Checked that it's not already running in task manager.
- Killing nearly other program running to see if there was conflicts.

###Log file

[INFO] 14:11:19: 
========================================
Operation System: Windows 10 Pro
System Architecture: 64 bit
Environment OS Version: Microsoft Windows NT 6.2.9200.0
System Directory: C:\WINDOWS\system32
Executing Directory: C:\Aurora
Launch Directory: C:\Aurora
Processor Count: 8
SystemPageSize: 4096
Environment Version: 4.0.30319.42000
Is Elevated: True
Aurora Assembly Version: 1.0.0.0
Aurora File Version: 0.8.1
========================================

[INFO] 14:11:20: [ASUS] Could not find Asus Aura SDK, please install version v1.07.71
[INFO] 14:11:20: Loading Configuration
[INFO] 14:11:20: Loading Plugins
[INFO] 14:11:20: Loading KB Layouts
[INFO] 14:11:20: Loading Input Hooking
[INFO] 14:11:20: Loading RazerSdkManager
[FATAL] 14:11:20: RazerSdkManager failed to load!
[FATAL] 14:11:20: System.Threading.WaitHandleCannotBeOpenedException: No handle of the given name exists.
   at System.Threading.EventWaitHandle.OpenExisting(String name, EventWaitHandleRights rights)
   at RazerSdkWrapper.Utils.EventWaitHandleHelper.Pulse(String name)
   at RazerSdkWrapper.RzSdkManager..ctor()
   at Aurora.App.OnStartup(StartupEventArgs e) in C:\projects\aurora\Project-Aurora\Project-Aurora\App.xaml.cs:line 334
[INFO] 14:11:20: Loading Applications
[INFO] 14:11:23: Fetching latest pointers
[INFO] 14:11:23: Loading Device Manager
[INFO] 14:11:23: Loading LGS DLL
[...]

PR #2324, opened 2021-06-16
Add configuration option to specify custom port for GameStateListener

SheepGoMeh:
Offers a possible fix/workaround for #2300

- Ability to specify a custom port for GameStateListener for cases where the port might be used or the port range can't be used for whatever reason


diogotr7:
I won't merge this just because all the game mods assume this port. This would break them all and i'm not going to update them to account for this. Sorry! Maybe working without the server running would be a better approach?

